PR #118, opened 2023-05-02
Multiple statistics-functions in montecarlo.estimate()

pnkraemer:


pnkraemer:
Don't merge. Split the "new" function into "estimate" (old behaviour -- no changes necessary in hutch.py" and "multiestimate" new function, for now it is hidden.

PR #229, opened 2025-01-07
Automatically implement vector matrix products with `jax.vjp`

### LINKED ISSUE #228: Use jax.linear_transpose to implement vA
Split out of #226

pnkraemer:
Resolves #228.

**This change breaks the current API** because all bidiagonalisation-based algorithms no longer require vector-matrix products to be passed to the decomposition. This includes:

- `decomp.bidiag`
- `eig.svd_partial`
- `funm.integrand_funm_product`

Instead of requiring the vector-Jacobian product, all functions now use `jax.vjp` to turn a matrix-vector product into a vector-matrix product. We don't use `jax.linear_transpose` because the forward pass is required for bidiagonalisation anyway, so we can do evaluation+transposition in a single call to `vjp` instead of a call to, each, forward pass and transposition.

The next version number will reflect the backwards-incompatible change.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ default_language_version:
   python: python3
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.6.0
+    rev: v5.0.0
     hooks:
       - id: check-yaml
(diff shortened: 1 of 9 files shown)


tjltjl:
Nice!

PR #234, opened 2025-01-07
Implement `eig.eig_partial`

### LINKED ISSUE #227: Implement partial eigenvalue decompositions
This came out of #225.  Concretely: build `matfree.eig.eig_partial` by combining Hessenberg factorisations (Arnoldi) with `jnp.linalg.eig`, and perhaps also build `matfree.eig.eigh_partial` by combining tridiagonalisation (Lanczos) with `jnp.linalg.eigh`. 

Progress: 

- [x] `eig_partial` (See #234)
- [x] `eigh_partial` (See #233)

pnkraemer:
Like `eigh_partial` and `svd_partial`.

Resolves #227.

### DIFF
--- a/matfree/eig.py
+++ b/matfree/eig.py
@@ -1,10 +1,10 @@
 """Matrix-free eigenvalue and singular-value analysis."""
 
-from matfree.backend import linalg
+from matfree.backend import linalg, np
 from matfree.backend.typing import Array, Callable
 
 
-def svd_partial(bidiag: Callable):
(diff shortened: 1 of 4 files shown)


qiyang-ustc:
I made some simple tests, it seems the forward works well but the AD does not work at all. Maybe we should specify the correct AD rule or other attempts to make it simple to use.

pnkraemer:
Thanks for the update! How do you feel about turning this into a separate issue so we can figure out the best solution there?

PR #247, opened 2025-07-27
Correct variable name in loop.

Whatisthisname:
Groundbreaking PR. You can also just fix do it yourself without the PR if you agree, I just got stuck on it for longer than I would have liked 😖


pnkraemer (approved):
Thanks!

PR #248, opened 2025-08-01
Rename variable to adhere to paper

Whatisthisname:
Update `eta` to `gamma` and lowercase "H" in `dH_k`.


pnkraemer (approved):
Thanks! 

(If someone sees this, the reason for this renaming these variables is so that the variable names in code match those in [maths](https://arxiv.org/pdf/2405.17277) .)

PR #263, opened 2026-05-18
feat(stochtrace): Add exchangeable LOO-based estimators of the trace

### LINKED ISSUE #14: Implement the `XTrace` estimator
https://arxiv.org/pdf/2301.07825.pdf sounds promising

Also explains some other approaches for variance reduction.

sethaxen:
As suggested in #14, this PR adds to `stochtrace` an estimator and integrands for estimating the trace ~, diagonal, and squared row norm~ as described in various algorithms in https://arxiv.org/abs/2512.15929:
- [x] `leave_one_out_xtrace` (XTrace)
- [x] `leave_one_out_xnystrace` (XNysTrace)
- ~[ ] `leave_one_out_xdiag` (XDiag)~
- ~[ ] `leave_one_out_xnysdiag` (XNysDiag)~
- ~[ ] `leave_one_out_xrownorm` (XRowNorm)~

Closes #14


sethaxen:
Currently only XTrace is implemented in order to work out any major API/approach issues before continuing. Some notes for consideration:

- It might be cleaner to change the API to e.g. `integrand_trace_loo(method: Literal["svd", "nystrom"] = "svd", resphere=True)` to keep fewer API methods. It's a little annoying that the method would be randomized SVD when we actually only use the first half of the randomized SVD which actually doesn't even compute an SVD.
- If `num_samples` > `n`, then the XTrace implementation will fail, since it expects `R` to be square upper triangular. However, if the `samples` are full-rank, then providing more than `n` samples is wasteful, since we can then compute the trace exactly. Or we could even then replace the `samples` with the identity matrix to make it deterministic. Is this a case we need to detect and work around in practice?
- The XTrace discussion in the paper and thesis assumes that `Y = vmap(matvec)(samples).T` is full-rank and will produce `nan`s if it is not (because computing the downdate matrix `S` requires inverting `R`). There are really just 2 cases where it wouldn't be:
    + sometimes if `samples` is rank-deficient. This almost never happens for normal test vectors, but for Rademacher and small `n`, this can indeed happen. Is this a case we need to handle?
    + If the rank of the operator `B` is lower than the `num_samples`. This is quite a serious problem actually, because in practice a user may not even know the rank of their operator. If we can assume `samples` is full-rank, then `rank(Y) = rank(B)`, and we show that the LOO residual is 0 for all samples, so that the exact trace can be computed and returned with less computation. However, if a user provided Rademacher samples, then this assumption would be false. There is a way to modify the algorithm to handle singular `R`, but it requires computing the SVD of `R`. However, this would be more work. Should we just assume `samples` is full-rank? (what the code currently does)
- I implemented the method and tested it with complex support just in case that's something you want to support here.

matfree/stochtrace.py
```diff
@@ -35,6 +35,138 @@ def estimate(matvecs, key, *parameters):
...
+        SIAM J Matrix Anal A. 45.1: 1-23.
+        doi: [10.1137/23M1548323](https://doi.org/10.1137/23M1548323)
+        arXiv: [2301.07825](https://arxiv.org/abs/2301.07825)
+    - Epperly EN (2025). Make the most of what you have: Resource-efficient randomized algorithms for matrix computations. PhD Thesis.
+        arXiv: [2512.15929](https://arxiv.org/abs/2512.15929)
+    """
+
+    # TODO: handle case where num_samples > n (violates thin QR constraint)
+    def integrand(matvec, samples, *params):
+        Omega, unflattens = func.vmap(tree.ravel_pytree)(samples)
```
sethaxen: Looks a bit clumsy. Do you think this would do the right thing for arbitrary `samples`?

matfree/stochtrace.py
```diff
@@ -35,6 +35,138 @@ def estimate(matvecs, key, *parameters):
...
+            out_axes=-1,
+        )
+
+        Y = matmat(Omega, unflattens)
+        Q, R = linalg.qr_reduced(Y)
+        Z = matmat(Q, unflattens)
+
+        def _trace_exact():
+            tr_B = linalg.vdot(Q, Z)
+            return np.ones((num_samples,), dtype=R.dtype) * tr_B
```
sethaxen, line 140: Maybe should be `np.full` if I add that to the backend.

matfree/stochtrace.py
```diff
@@ -35,6 +35,138 @@ def estimate(matvecs, key, *parameters):
...
+    vectors, this variant passes the **full** test matrix to the integrand at
+    once.  This is required for the XTrace family of algorithms, which need a
+    batch QR or Cholesky decomposition before the second matvec phase.
+
+    Parameters
+    ----------
+    integrand
+        An integrand that accepts ``(matvec, Omega, *parameters)`` where
+        ``Omega`` has shape ``(num, n)``.  The integrand is responsible for
+        averaging over the sample axis internally.
```
pnkraemer: What does the docstring mean by "averaging over the sample axis"?

matfree/stochtrace.py
pnkraemer: I am a bit surprised this even works, to be honest... I personally would've gone the slow way by reading "unflatten" from a single sample and then vmapping over lambda s: ravel_pytree(s)[0] to get Omega. I am not saying this is the only way to do this, though :)

matfree/stochtrace.py
```diff
@@ -35,6 +35,138 @@ def estimate(matvecs, key, *parameters):
     return estimate
 
 
+def estimator_loo(integrand: Callable, /, sampler: Callable) -> Callable:
```
pnkraemer: What's your opinion on "estimator_loo" vs "estimator_leave_one_out"? I understand that LOO is a common acronym, but I wonder whether being explicit here would make the implementation age better.

matfree/stochtrace.py
```diff
@@ -35,6 +35,138 @@ def estimate(matvecs, key, *parameters):
...
+    """
+
+    def estimate(matvec, key, *parameters):
+        samples = sampler(key)
+        return np.mean(integrand(matvec, samples, *parameters), axis=0)
+
+    return estimate
+
+
+def integrand_trace_svd(*, resphere: bool = True) -> Callable:
```
pnkraemer: I think it would be ideal if we could distinguish the LOO integrands from the other integrands. Otherwise, it might be tempting to mix them. For example, we could use a naming pattern along the lines of `leave_one_out_xtrace` (respectively `leave_one_out_xdiag`, leave_one_out_xnystrace`, etc.). Thoughts? 

Similarly, I am open to renaming the existing `estimator` to `estimator_monte_carlo` and the integrands to `monte_carlo_trace` etc. But that is somewhat independent of this PR.

matfree/stochtrace.py
```diff
@@ -35,6 +35,138 @@ def estimate(matvecs, key, *parameters):
...
+    Returns
+    -------
+    estimate
+        A function ``estimate(matvec, key, *parameters) -> result``.
+
+    """
+
+    def estimate(matvec, key, *parameters):
+        samples = sampler(key)
+        return np.mean(integrand(matvec, samples, *parameters), axis=0)
```
pnkraemer, line 73: Does this implement the averaging over "m", respectively "m/2" sub-trace-estimators? E.g., Equation 1.9 in the paper.

Is it possible to also move, for example, the matmat + QR logic in here? It seems to be reused across trace, diagonal, and row norms, and by moving the common code here, we can handle corner cases such as the number of samples exceeding the matrix size once and for all. Thoughts?

matfree/stochtrace.py
pnkraemer: Btw, is there a specific reason for naming the flag "resphere"? I typically like verbose names for boolean flags, e.g., `apply_resphering` or so, but if there is a reason to choose that variable name, then all good 👍

matfree/stochtrace.py
```diff
@@ -35,6 +35,138 @@ def estimate(matvecs, key, *parameters):
...
+                + linalg.vecdot(X, H @ X, axis=0)
+                + W_vd_S * linalg.vecdot(S, R, axis=0)
+            )
+            return tr_B_hat_loo + residual_scale * tr_residual_loo
+
+        Y_rank = np.sum(np.abs(linalg.diagonal(R)) > np.finfo_eps(R.dtype))
+
+        # NOTE: assumes rank(Y) == rank(B), which is almost always true
+        # for Gaussian samples and, when n is large enough, for Rademacher samples.
+        return control_flow.cond(
```
pnkraemer: I would like to make sure I understand this line: We compute the thin QR of `Y = A @ Omega`, which has the shape (n, num_samples). 
- For the QR logic to be valid, we need n >= num_samples. Right?
- For the estimator to be valid, we additionally need that R is invertible, which means that Y has full rank (i.e., rank `num_samples`).

Does that mean that if n < num_samples, we should raise an exception, ignore/remove excessive samples (then the estimate computes the exact solution anyway, right?), or do something else, for example, materialise A and evaluate the trace exactly, which would be cheaper than doing xtrace with excessive samples. What do you think?  I am leaning towards raising an exception (I do the same in Arnoldi, see line 420 in decomp.py), especially if the QR isn't well-defined anymore.

If Y is a tall matrix (n >= num_samples) but rank-deficient (rank < num_samples), then the linear solve fails because R isn't invertible? Do I read correctly that in this case, we evaluate the exact trace? Or is the name "exact trace" slightly inaccurate here? 

Which case does the Y_rank == n represent? 

(Sorry for the long rant, but I am trying to understand some of the assumptions behind this code)

tests/test_stochtrace/test_loo_trace_svd.py
```diff
@@ -0,0 +1,79 @@
+"""Tests for integrand_trace_svd (XTrace)."""
+
+import jax
+import jax.numpy as jnp
+import pytest
+from matfree import stochtrace
```
pnkraemer: I understand this is a draft, but for completeness, I use `matfree.backend.*` in the tests as well. I only juse jax directly in the tutorials.

tests/test_stochtrace/test_loo_trace_svd.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+jax.config.update("jax_enable_x64", True)
```
pnkraemer: If my understanding is correct, this line makes all of matfree's tests run in double precision. Instead of changing this globally, I typically set it to True at the beginning of a test and to false again afterwards. (See eg the Hessenberg-adjoint tests for an example)

tests/test_stochtrace/test_loo_trace_svd.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+@pytest.mark.parametrize("resphere", [True, False])
+@pytest.mark.parametrize("dtype", [jnp.float64, jnp.complex128])
```
pnkraemer: I really like the idea of supporting complex arithmetic! If we replace the `jnp.float64` with Python's `float`, and `jnp.complex128` with `complex`, then the tests also run in single precision (respectively in double precision if we enable it).

sethaxen:
Figure 4 of the paper makes a good case that for test vectors, all choices are equivalent for operators with flat spectral decay, while for operators with flat spectra resphered(normalized) >> sphere > sign >> normal:
Screenshot_2026-05-19_09-22-13

Resphering works with normal or sphere vectors. Based on this, I wonder if it makes sense to add a `sampler_sphere` method and recommend its use in the docstrings for `estimator_loo`. [This post](https://www.ethanepperly.com/index.php/2024/01/28/dont-use-gaussians-in-stochastic-trace-estimation/) also argues that even for plain Hutchinson estimators, normal distributions are worse than Rademacher and sphere, which each have their trade-offs.

tests/test_stochtrace/test_loo_trace_svd.py
```diff
@@ -0,0 +1,79 @@
...
+    sampler = stochtrace.sampler_normal(jnp.ones(n, dtype=dtype), num=35)
+    integrand = stochtrace.integrand_trace_svd(resphere=resphere)
+    estimate = stochtrace.estimator_loo(integrand, sampler)
+    matvec = lambda v: U @ (d * (U.T.conj() @ v))
+
+    key_ests = jax.random.split(key, num_rep)
+    received = jax.vmap(lambda key: estimate(matvec, key))(key_ests)
+    rel_err = jnp.abs(received - expected) / jnp.abs(expected)
+    mean_rel_err = jnp.mean(rel_err)
+    assert float(mean_rel_err) < 1e-5
```
pnkraemer: Do I understand correctly that this block essentially runs the test for multiple seeds at once? If so, very nice! I typically parametrise the tests with the seed, but I think your version should be a lot faster. 

Some thoughts on this test:
- How about we test that jnp.amax(rel_error) < 1e-5 instead of the mean? If the errors are on different orders of magnitude, the mean distorts them.
- Can we make the matvec parametrised (as in `def matvec(v, U_and_d): ...` and call `estimate(matvec, key, U, d)` to test the parametrisation as well?

pnkraemer (commented):
Thanks for opening this draft PR! I left some comments in the diff, mainly to make sure I understand the maths well enough, but also some code design questions. 

Regarding the fewer API methods: How about we let this depend on how much code is reused? For example, if we move more logic to estimator_loo, then maybe what's left of SVD vs Nyström is so different that a flag would obscure more than help, or maybe they still share almost all the code, in which case a flag would be nicer?

Regarding `num_samples > n` and full rank: I asked a closely related question in the diff; let's continue the discussion there?

Regarding complex support: really nice! I made a small suggestion in the diff regarding data types.

Regarding sampler_sphere: That would be awesome! If it can also be used for existing Hutchinson, should we do this in a separate PR?

Thanks again for pushing XTrace!

matfree/stochtrace.py
sethaxen: Oh, this is an outdated carryover from an earlier version where the `mean` over the LOO estimates was computed by the integrand. I'll fix.

matfree/stochtrace.py
sethaxen: I don't have a strong preference. `_loo` makes more sense if we end up including that in the integrand functions as well so the function names don't get too long, but explicit is also good.

matfree/stochtrace.py
sethaxen: > Does this implement the averaging over "m", respectively "m/2" sub-trace-estimators? E.g., Equation 1.9 in the paper.

Yes!

> Is it possible to also move, for example, the matmat + QR logic in here? It seems to be reused across trace, diagonal, and row norms, and by moving the common code here, we can handle corner cases such as the number of samples exceeding the matrix size once and for all. Thoughts?

This would work for XTrace and XDiag but not for XRowNorm. (For an operator B, XTrace and XDiag compute the first part of the randomized SVD for `B`, while XRowNorm does it for `B.T @ B`. But later in the algorithm it requires another product by `B` alone, so we can't just replace the matvec with its composition with an adjoint). But also, if we moved that logic in here, then we couldn't use it for the Nystrom algorithms.

matfree/stochtrace.py
sethaxen: > I think it would be ideal if we could distinguish the LOO integrands from the other integrands. Otherwise, it might be tempting to mix them. For example, we could use a naming pattern along the lines of leave_one_out_xtrace (respectively leave_one_out_xdiag, leave_one_out_xnystrace`, etc.). Thoughts?

I agree that the risk of misuse is high, and being more explicit with the names would be good to decrease the risk. The proposed renames would do that. I wonder if `leave_one_out_xtrace` is redundant since XTrace inherently is leave-one-out. One could alternatively just do `leave_one_out_trace`. BTW, the X in all of these algorithms stands for "exchangeable", which refers to the exchangeability coming from using LOO (as opposed to Hutchinson which is also exchangeable).

> Similarly, I am open to renaming the existing `estimator` to `estimator_monte_carlo` and the integrands to `monte_carlo_trace` etc. But that is somewhat independent of this PR.

That might make sense! I think the `estimator`/`integrand` naming makes sense when only Hutchinson-like estimators are supported, but it doesn't seem to generalize well. A downside to dropping `integrand` is that it makes it unclear which functions are accepted by an `estimator` method.

An alternative would be to rename the `stochtrace` module to `hutchinson` and put the LOO-based estimators somewhere else? Or keep it as `stochtrace` and add e.g. `xstochtrace`? That also would solve the problem of users accidentally mixing "integrands" and estimators.

matfree/stochtrace.py
sethaxen: I _think_ this works because the callable returned by `ravel_pytree` is itself a pytree? (It's a [HashablePartial](https://github.com/jax-ml/jax/blob/b53699b63ea40db643ffb1a9d9c27974beb331c6/jax/_src/util.py#L596-L612)): But also, this isn't documented behavior, so maybe your approach would be safer.

matfree/stochtrace.py
sethaxen: > I would like to make sure I understand this line:...

Yes, all correct.

I think I prefer:
- Raise an exception for `num_samples > n`
- For `num_samples == n`, materialize `A` and compute its exact trace (cheaper than working with `samples` and guaranteed to be exact even if `samples` is rank-deficient

> If Y is a tall matrix (n >= num_samples) but rank-deficient (rank < num_samples), then the linear solve fails because R isn't invertible? Do I read correctly that in this case, we evaluate the exact trace? Or is the name "exact trace" slightly inaccurate here?

So there are two reasons `Y` might be rank-deficient. It could be because `A` is rank-deficient or because `samples` is rank-deficient, or both. If `samples` is full-rank, then it can only be due to `A`. Then `Y` being rank-deficient implies that the low-rank approximation to `A` that XTrace implicitly constructs (`Q @ Q.H @ A`) is actually not an approximation at all but is exact (i.e. within floating point error). Furthermore, since we have `rank(Y) < rank(B) <= num_samples`, if we now leave out a row from `samples`, we get `rank(Y_i) <= rank(B) <= num_samples`, so all of the leave-one-out residuals are 0 as well, and we can skip all of the downdate logic and just return the exact trace.

However, if `Y` is rank-deficient at least in part due to `samples` being rank-deficient, then the residuals are not necessarily 0 and we need a different solution to compute the downdate. I'm still nailing down some of the math, but I believe what we need to do is compute the downdate `S` using the pseudoinverse, setting a column to 0 if the corresponding entry of `diag(pinv(R)@R)` is not 1. This requires an additional SVD of R, which isn't ideal but for `num_samples << n` has negligible cost. I'm investigating whether we could replace the QR with a rank-revealing (pivoted) QR (i.e. `jax.scipy.linalg.qr` with `pivoting`), in which case the pseudoinverse can be computed with a triangular solve.

`Y` would only be rank-deficient if the user used Rademacher samples and `n` was small. The code complexity would be lower if we could just assume `Y` is full-rank. So one solution is to jettison any extra complexity here and just document that the implementation assumes full-rank `samples` and recommend `sphere`.

> Which case does the Y_rank == n represent?

This is handling the case where `Y` is full-rank and `num_samples == n` (`num_samples > n` still errors when tracing the other branch). If we've already special-cased `num_samples == n`, then this check isn't necessary.

tests/test_stochtrace/test_loo_trace_svd.py
sethaxen: Agreed! Currently this is explicit because the tolerances are actually taken from experiments in the paper, which were computed using 64-bit precision. Any tolerances I could choose for 32-bit precision would be arbitrary.

tests/test_stochtrace/test_loo_trace_svd.py
sethaxen: > Do I understand correctly that this block essentially runs the test for multiple seeds at once?

> How about we test that jnp.amax(rel_error) < 1e-5 instead of the mean? If the errors are on different orders of magnitude, the mean distorts them.

Ah, so this is actually replicating experiments from the paper (specifically the `step` and `exp` experiments), just with fewer replicates (10 instead of 1000). They report the mean relative error, and I'm testing that we see similar mean relative error. I actually don't know what bounds we should expect on each relative error, so any threshold I could choose there would be arbitrary.

> Can we make the matvec parametrised (as in def matvec(v, U_and_d): ... and call estimate(matvec, key, U, d) to test the parametrisation as well?

Ah, good idea!

sethaxen:
> Regarding sampler_sphere: That would be awesome! If it can also be used for existing Hutchinson, should we do this in a separate PR?

Sure, I can do it separately before this one is merged, because it might make sense for this PR to include in the docstring a recommendation to use it.
 
> Thanks again for pushing XTrace!

You're welcome!

matfree/stochtrace.py
pnkraemer: I pretty much agree with everything you say! 

Regarding separate modules, I lean towards having xtrace and Hutchinson in the same module, because they solve related problems and share some code (eg samplers). How about we simply drop the term "integrand" then for all methods? I am not even sure "integrand" is a suitable name for the LOO estimators because we don't really do Monte Carlo anymore. I like your suggestion of `leave_one_out_{trace,diag,rownorm,...}`, which could be realised as:
```python
def estimator_leave_one_out(leave_one_out, sampler):
    ....sampling and averaging code goes here...

sample = sampler_rademacher(...)
loo = leave_one_out_trace(...)
estimate = estimator_leave_one_out(loo, sample)
trace = jax.jit(estimate)(matvec, ...)
```
and, later, with downstream edits in the Hutchinson code like:
```python
def estimator_monte_carlo(monte_carlo, sampler):  # currently estimator()
    ....sampling and averaging code goes here...

sample = sampler_rademacher(...)
mc = monte_carlo_trace(...)  # currently integrand_trace()
estimate = estimator_monte_carlo(mc, sample)
trace = jax.jit(estimate)(matvec, ...)
```

What do you think of this API?

matfree/stochtrace.py
pnkraemer: > I think I prefer:

Sounds good! I like this distinction because it doesn't depend on any runtime values because shapes are known at compile time. Thus, we can use standard "if-else" logic and so on, which will make the code nicely readable.

> Y would only be rank-deficient if the user used Rademacher samples and n was small. The code complexity would be lower if we could just assume Y is full-rank. So one solution is to jettison any extra complexity here and just document that the implementation assumes full-rank samples and recommend sphere.

Thanks for explaining! Since for tiny matrices, using stochastic trace estimation is a bit pointless anyway, I would be okay with assuming that the sample-matrix is full rank; at least for now. If we notice we need to handle this low-rank case more precisely, we can always iterate. But I'd like to avoid making the code significantly more complicated to solve a problem that we don't have. Maybe we can state this assumption in the docstring somewhere to be transparent (including a mentioning of the small-matrix-rademacher case). 


So we retain the control_flow.cond(Y_rank < n_samples, trace_exact, trace_estimate)? Can we add a brief explanation in the source code along the lines of what you posted above (or a more compact version thereof, I don't have a preference)

tests/test_stochtrace/test_loo_trace_svd.py
pnkraemer: Ah nice, in this case, keep the double precision and the mean of relative errors as is, but perhaps add a small comment that this test reproduces the paper experiments? (So I/we know not to mess with it much down the line)

matfree/stochtrace.py
pnkraemer: I see. That's a shame... It is a bit unsatisfying how similar the new estimator looks (in comparison to the current estimator): the only difference is vmapping over the samples vs not doing it. But I suppose sometimes we have to live with unsatisfying code... :) 

(btw, should we do a tree_map around the mean? Does the current implementation handle pytree-valued linear operators? See the tests in https://github.com/pnkraemer/matfree/blob/main/tests/test_funm/test_integrand_funm_product_logdet.py for an example of how we verify this for matrix functions)

matfree/stochtrace.py
pnkraemer: That's my thinking as well. Wrapping the callable returned by pytree-ravelling into a Pytree itself works for things like jit, eval_shape, and so on, but I am quite surprised it works for vmap as well! So yeah, I think my approach would be safer (in the sense of more obviously correct), but I am also okay with anything that passes the tests reliably.

matfree/stochtrace.py
pnkraemer: Sure, add it if you like! Though this is a one-liner already without adding a function to the backend, so I don't think having it in the backend is critical.

matfree/backend/linalg.py
```diff
@@ -52,6 +58,14 @@ def inner(x1, x2, /):
     return jnp.inner(x1, x2)
 
 
+def vecdot(x1, x2, /, axis=-1):
```
pnkraemer: I've just seen this: the backend now has a `vecdot`, a `vdot`, and an `inner`. I know numpy and jax.numpy have these different options, but one goal of channelling the array computations through matfree.backend.linalg was to avoid this kind of redundancy. Can we eliminate one of these inner products easily? If not, we can keep them.

sethaxen: Yeah it's annoying and I always need to look up which is the correct one to use. What we're missing with `inner` is something that can compute the Frobenius inner product between two matrices or the Frobenius inner products of their column pairs.

`vecdot(x1, x2, axis=-1)` could be dropped and instead use `func.vmap(linalg.vdot)(x1, x2, in_axes=0)` or even `func.map(linalg.inner)(x1.conj(), x2)`.
`vdot(x1, x2)` can be replaced with `linalg.inner(reshape(x1.conj(), shape=-1), reshape(x2, shape=-1))`, it's just uglier.

IMO the `vdot` replacement is ugly and harder to read. I lean toward keeping `vdot` and, since `vecdot`'s replacement using `vdot` is still quite elegant, dropping `vecdot`. Thoughts?

matfree/stochtrace.py
sethaxen: > What do you think of this API?

Looks good to me, although there's still the issue of how to distinguish xtrace vs xnystrace. Actually this is a good argument for your earlier idea of `leave_one_out_xtrace`; alternatively, `leave_one_out_trace(resphere: bool, method)` where `method` could be either `Literal["svd", "nystrom"]` for the approximation used or `Literal["xtrace", "xnystrace"]` for the algorithm used. Yet another alternative would be a `assume_psd` boolean argument, but I don't much care for this one, since xnystrace isn't *universally* better than xtrace even for PSD matrices.

sethaxen: > Btw, is there a specific reason for naming the flag "resphere"? I typically like verbose names for boolean flags, e.g., `apply_resphering` or so, but if there is a reason to choose that variable name, then all good 👍

Sorry, I missed this one earlier. Epperly refers to this method as "normalization" in the XTrace paper but adopts the resphering terminology for the thesis because it's more descriptive.

The name is presumably because after applying resphering, the test vectors are spherically sampled in the subspace containing the residual, so if they were already spherically sampled ("sphered", or Gaussian), they are now resphered. I agree it's more descriptive than normalization. He refers to the estimators that use resphering as "resphered", but I'm not opposed to switching to `apply_resphering`.

pnkraemer: My original motivation for leave_one_out_xtrace was that the algorithm is known as "xtrace", so that makes it easy to connect. In this line of thinking, the others would be "xnystrace", "xdiag", etc., which then mirrors the names given by the original author. So I think that having "leave_one_out_xtrace", "leave_one_out_xnystrace", "leave_one_out_xdiag", and so on would be my favourite. But I don't have a strong opinion here.

matfree/backend/linalg.py
pnkraemer: Sounds good. Let's keep vdot then and if we want to map along an axis, let's use vmap for that. I think that's the best combination of readable and easy to maintain. 

In the long run, one could remove `inner` because it implements vdot without conjugation (and we always want conjugation anyway). But that is a separate concern :+1:

matfree/stochtrace.py
pnkraemer: Regarding resphering: I didnt think of it as a verb. Now it makes more sense to me. I am okay with it as is, or we can switch to the more verbose option. When in doubt, I tend to prefer verbose variable names, but I am really okay with both here. 

Sorry for having two different discussions in the same thread...

matfree/stochtrace.py
sethaxen: I'll leave it out for now.

matfree/stochtrace.py
sethaxen: You know, I think we actually need to get the array of `unflattens`. The problem is that for an arbitrary pytree, we actually don't have a good way to get just the first sample. In order to index into it, we have to ravel the pytree first, but then we can't get back to the pytree without first having the `unflatten` callable.

sethaxen:
Alright, I think I've incorporated all suggestions into the xtrace implementation. Next I would open a PR with a `sampler_sphere` method, and then I would continue with the other implementations here along the same lines as the xtrace one.

matfree/stochtrace.py
pnkraemer: But we know that the samples that are passed to the integrand are batched along the zeroth axis, right? So I think the following might give us the first sample:
```python
from matfree.backend import tree
sample0 = tree.tree_map(lambda s: s[0], samples)
_sample0_flat, unflatten0 = tree.ravel_pytree(sample0)
```
Thoughts?

matfree/stochtrace.py
```diff
@@ -35,6 +35,153 @@ def estimate(matvecs, key, *parameters):
...
+        def matvec_flat(v, unflatten):
+            return tree.ravel_pytree(matvec(unflatten(v), *params))[0]
+
+        if num_samples == n:
+            # It's faster and more accurate to compute the trace exactly and deterministically
+            # when num_samples == n
+            Omega = np.eye(n, dtype=Omega.dtype)
+            B_diag = func.vmap(
+                lambda v, unflatten, i: matvec_flat(v, unflatten)[i], in_axes=-1
+            )(Omega, unflattens, np.arange(n))
```
pnkraemer: Equivalent but simpler (assuming the unflattening is absorbed into the matvec, and taking the Omega from above, no need to form an identity matrix):
```python
B_diag = jax.jacfwd(matvec_flat)(Omega)
```
which materialises the full matrix because the matvec is a linear operation.

matfree/stochtrace.py
```diff
@@ -35,6 +35,153 @@ def estimate(matvecs, key, *parameters):
...
+            return np.ones((num_samples,), dtype=R.dtype) * tr_B
+
+        def _trace_estimate():
+            S = linalg.solve_triangular(R, np.eye(R.shape[0], dtype=R.dtype), trans=2)
+            S = S / func.vmap(linalg.vector_norm, in_axes=-1)(S)
+
+            Q_H = Q.T.conj()
+            H = (
+                Q_H @ Z
+            )  # tr(H) == tr(B_hat), where B_hat = Q @ Q.H @ B is a low-rank approximation to the operator B
```
pnkraemer:
```suggestion
            # tr(H) == tr(B_hat), where B_hat = Q @ Q.H @ B is a low-rank approximation to the operator B
            H = Q_H @ Z 
```

matfree/stochtrace.py
```diff
@@ -35,6 +35,153 @@ def estimate(matvecs, key, *parameters):
...
+            W = Q_H @ Omega
+            T = Z.T.conj() @ Omega
+            W_vd_S = func.vmap(linalg.vdot, in_axes=1)(W, S)
+            X = (
+                W - S * W_vd_S.conj()
+            )  # samples.T projected onto the subspace spanned by Q_i, i.e. Q formed leaving out samples[i, :]
```
pnkraemer:
```suggestion
            # samples.T projected onto the subspace spanned by Q_i, i.e. Q formed leaving out samples[i, :]
            X =  W - S * W_vd_S.conj()
 ```

File: matfree/stochtrace.py [hunk 16]
```diff
@@ -35,6 +35,153 @@ def estimate(matvecs, key, *parameters):
...
+                    sqnorm_samples_projected == 0.0, 1.0, sqnorm_samples_projected
+                )
+                residual_scale = rank_residual / sqnorm_samples_projected
+            else:
+                residual_scale = 1.0
+
+            tr_B_hat = linalg.trace(H)
+            tr_B_hat_loo = tr_B_hat - func.vmap(linalg.vdot, in_axes=1)(
+                S, H @ S
+            )  # tr(B_hat) leaving out one sample
```
pnkraemer:
```suggestion
            # tr(B_hat) leaving out one sample
            tr_B_hat_loo = tr_B_hat - func.vmap(linalg.vdot, in_axes=1)(S, H @ S)  
```

tests/test_stochtrace/test_leave_one_out_xtrace.py
```diff
@@ -0,0 +1,145 @@
+"""Tests for leave_one_out_xtrace."""
+
+from matfree import stochtrace, test_util
+from matfree.backend import config, func, linalg, np, prng, testing
+
+config.update("jax_enable_x64", True)
```
pnkraemer: Can we change this setting inside the test (and deactivate it again at the end of the test)? At the toplevel, it will affect all other tests in the library.

tests/test_stochtrace/test_leave_one_out_xtrace.py
```diff
@@ -0,0 +1,145 @@
...
+        match=f"Number of samples num={n + 1} exceeds the acceptable range. Expected: 1 <= num <= {n}.",
+    ):
+        estimate(matvec, key, A)
+
+
+# reproduces the results of the experiment "exp" from the XTrace paper
+@testing.parametrize("apply_resphering", [True, False])
+@testing.parametrize("dtype", [float, complex])
+def test_xtrace_fast_spectral_decay(apply_resphering, dtype):
+    """Assert that the trace of a matrix with fast spectral decay is estimated accurately."""
```
pnkraemer:
```suggestion
@testing.parametrize("apply_resphering", [True, False])
@testing.parametrize("dtype", [float, complex])
def test_xtrace_fast_spectral_decay(apply_resphering, dtype):
    """Assert that the trace of a matrix with fast spectral decay is estimated accurately.

    Reproduces the results of the experiment 'exp' from the XTrace paper.
    """
```

tests/test_stochtrace/test_leave_one_out_xtrace.py
```diff
@@ -0,0 +1,145 @@
...
+    rel_err = np.abs(received - expected) / np.abs(expected)
+    mean_rel_err = np.mean(rel_err)
+    assert float(mean_rel_err) < 1e-5
+
+
+# reproduces the results of the experiment "step" from the XTrace paper
+@testing.parametrize("apply_resphering", [True, False])
+@testing.parametrize("dtype", [float, complex])
+def test_xtrace_large_spectral_drop(apply_resphering, dtype):
+    """Assert that the trace of a matrix with some large eigenvalues and the rest small is estimated accurately."""
```
pnkraemer:
```suggestion
@testing.parametrize("apply_resphering", [True, False])
@testing.parametrize("dtype", [float, complex])
def test_xtrace_large_spectral_drop(apply_resphering, dtype):
    """Assert that the trace of a matrix with some large eigenvalues and the rest small is estimated accurately.
    
    Reproduces the results of the experiment 'step' from the XTrace paper.
    """
```

tests/test_stochtrace/test_leave_one_out_xtrace.py
```diff
@@ -0,0 +1,145 @@
...
+        {"fx": np.ones(n1), "fy": np.ones(n2)}, num=n1 + n2 - 1
+    )
+    estimate = stochtrace.estimator_leave_one_out(integrand, sampler)
+
+    received = estimate(matvec, key_est, A, B)
+    expected = linalg.trace(A) + linalg.trace(B)
+    assert np.allclose(received, expected, rtol=1e-2)
+
+
+config.update("jax_enable_x64", False)
```
pnkraemer: Ah, now I understand. What I meant in my last review with turning double precision off is that the config should be changed inside each test that needs it. I see how my comment wasn't precise enough. But turning it on and off at the toplevel doesn't have the desired effect, I think.

(That said, if the tests pass for you locally, I wonder whether double precision is even needed because I wouldn't be surprised if it isn't used this way.)

pnkraemer (commented):
Thanks again! Looks good (except for the test bit, where I think my last round of comments wasn't precise enough - sorry about that).

tests/test_stochtrace/test_leave_one_out_xtrace.py
sethaxen: Indeed it seems it wasn't necessary anymore! I've removed.

tests/test_stochtrace/test_leave_one_out_xtrace.py
sethaxen: It's now removed, see below comment.

matfree/stochtrace.py
sethaxen: I've implemented a version of this (was important to pass a single sample, and the result is the entire matrix, not just the diag), which required some tweaking to make sure if the user does use a sampler with a real dtype with a complex operator (or vice versa), then JAX doesn't error.

matfree/stochtrace.py
sethaxen: Yes, this worked!

matfree/stochtrace.py
```diff
@@ -45,6 +45,149 @@ def estimate(matvecs, key, *parameters):
...
+        Omega = func.vmap(lambda s: tree.ravel_pytree(s)[0])(samples).T
+        n, num_samples = Omega.shape
+
+        if num_samples > n:
+            raise ValueError(_error_num_samples(num_samples, maxval=n, minval=1))
+
+        def matvec_flat(v):
+            return tree.ravel_pytree(matvec(unflatten(v), *params))[0]
+
+        if num_samples == n:
```
sethaxen: The estimator performs `2*num_samples` matvecs and  more than `3*n*num_samples` allocations, whereas materializing the full operator requires `n` matvecs and `n**2` allocations. So should we take this branch if `2 * num_samples >= n`?

tests/test_stochtrace/test_leave_one_out_xtrace.py
pnkraemer: Awesome!

matfree/stochtrace.py
pnkraemer: Oh, good point. Are there possible allocation-downsides if `2 * num_samples < n < 3 num_samples` holds? If not, I think this is a great idea (especially if we document this rationale somewhere, e.g. in a source-code comment)!

matfree/stochtrace.py
```diff
@@ -137,6 +280,15 @@ def moment_fun(x, /):
     return integrand_wrapped
 
 
+def _materialize_operator(matvec_flat, x):
+    """Materialize the operator defined by an already-flattened matvec and a vector."""
+    # if x is real but the operator is complex, then jacfwd will error, so we perform
+    # a single matvec to find out if the result is complex
+    x = x.astype(matvec_flat(x).dtype)
```
pnkraemer: Two questions:
- Just to make sure: we shouldn't warn/error if the datatypes of operators and input samples mismatch? Numpy matvecs promote automatically, so if that is the baseline, then fixing things should be the right thing to do.
- Have you considered: `Ax_like = funv.eval_shape(matvec_flat, x); x = x.astype(Ax_like.dtype)`? This would achieve the same as the matvec but without any FLOPS I think.

sethaxen: > Just to make sure: we shouldn't warn/error if the datatypes of operators and input samples mismatch? Numpy matvecs promote automatically, so if that is the baseline, then fixing things should be the right thing to do.

Yeah mathematically, it should be fine. The estimators are all unbiased for any isotropic test vector (i.e. `E[w * w']=I), so there's no problem mathematically using a complex test vector with real or real with complex. And from that perspective I think we should allow it (and as we do, recommend using real with real and complex with complex). The only weirdness if using complex test vectors with real operator would be that one would get a trace estimate with a complex dtype back. I hardly think this would be a surprise though.

> Have you considered: `Ax_like = funv.eval_shape(matvec_flat, x); x = x.astype(Ax_like.dtype)`? This would achieve the same as the matvec but without any FLOPS I think.

Ah, great! That's much better.

matfree/stochtrace.py
sethaxen: > Oh, good point. Are there possible allocation-downsides if `2 * num_samples < n < 3 num_samples` holds?

No, there's a clear win when the number of samples is greater than half of the dimension.

Here's the breakdown. In terms of flops and allocs, with `m=num_samples`
- exact: `n` matvecs, `n^2` allocs
- estimate: `2m` matvecs, `3mn + 4m^2` allocs

So we can divide into 3 regimes:
- `m <= n/4`: estimate is better in terms of matvecs and allocations
- `n/4 <= m < n/2`: estimate is better in terms of matvecs but not allocations
- `m >= n/2`: exact is better in terms of both matvecs and allocations

I think we care more about keeping matvecs low than allocations low anyways, so the more conservative `m>=n/2` threshold seems the better one. Besides, it's always possible that XLA's optimizations will modify the number of actual allocations made, so it's hard to make concrete decisions based on those.

matfree/stochtrace.py
```diff
@@ -45,6 +45,291 @@ def estimate(matvecs, key, *parameters):
...
+    -------
+    nystrom_factor
+        A matrix `F` such that `F @ F.T.conj()` approximates the operator.
+    chol_upper
+    shift
+    """
+    n = Omega.shape[0]
+    Y = func.vmap(matvec_flat, in_axes=-1, out_axes=-1)(Omega)
+    Y_norm = linalg.vector_norm(Y)
+    shift = np.finfo_eps(Y.dtype) * Y_norm / n**0.5
```
sethaxen: I ran some tests with low-rank (rank < num_samples) PSD operators, where this shift from the paper would be critical for use of the Cholesky decomposition, and I found plenty of cases where this shift seemed too small for Cholesky to succeed. I could make tests pass by scaling this up by an arbitrary factor (e.g. 32). However, I found if I scaled too much, then I started to see the tests that reproduce the paper experiments fail (i.e. we don't get as low mean relative error as they saw in the paper).

I'm not sure yet the best way to handle this. Epperly's thesis does not justify this choice, only noting that it differs from that uses in an earlier paper by Tropp. I tried Tropp's, and it seemed too large. Tropp's paper also did not justify their choice and instead said it was inspired by an earlier paper, which, you guessed it, did not justify their choice.

Ideally we would like to find the lowest possible shift such that the Cholesky decomposition will succeed with high probability. Might be worth taking some time to think of a principled choice.

Alternatively, it might be the case as we saw with XTrace that if `rank <= num_samples` for full-rank `Omega`, then the Nystrom approximation is exact and the residual is 0. Then it would make sense to construct the Nystrom outer product form a different way, e.g. using the Hermitian eigendecomposition `H=UDU'` and returning the factor `F=Y@U / pinv(D)` where `pinv` sets all non-positive elements to 0 and inverts the rest. EDIT: The problem with using Cholesky failure to detect rank-deficiency is that I'm not certain that's guaranteed to be the reason.

sethaxen (commented):
I pushed an initial implementation of XNysTrace. I'm traveling next week so will continue polishing the following week when I get back. There's one serious open implementation question.

matfree/stochtrace.py
pnkraemer: Ah, I see. Thanks for explaining! So we have to decide what to do if `n/4 < m < n/2` holds, because there the estimate has better matvecs but worse allocations, whereas in all other cases there is a clear winner.

I am unsure whether we should optimise for matvecs or allocations, as it depends on how the code is used. Instead, I lean towards calling `estimate` instead of `exact` in all settings where there is no obvious winner, because that is what the user expects to happen when calling this function anyway. Fortunately, we arrive at the same solution anyway :) Calling `estimate` in the "in between" case?

matfree/stochtrace.py
pnkraemer: I see, thanks for explaining! Do I understand correctly that we would like the shift to be as small as possible? If so, how about making this an argument of the function with a suitable default, e.g., zero or machine epsilon, and mentioning in the docstring that this value might require tuning? 

More generally, have you considered replacing Cholesky + tridiagonal solve with alternative (pseudo)inverses, e.g., one using eigendecompositions? (I can elaborate if you like) The complexity wouldn't be much higher (IIRC, eigh is pretty fast on GPUs) and we could drop the shifting. I usually like to avoid shifts wherever possible. While using eigh instead of Cholesky deviates from the template implementation in the paper, I suspect it could be an improvement.

sethaxen: I tend to agree that eigh is preferable to arbitrarily shifting for Cholesky. The only complication is that eigh becomes non-differentiable with repeated eigenvalues, which would include if the rank of the operator is less than or equal to `num_samples-2`. There is a workaround that will likely make it into Jax itself (https://github.com/jax-ml/jax/issues/669#issuecomment-3079437240) and which I've been using for pretty much the exact same application (computing the pseudoinverse in a Nyström approximation) at https://github.com/sethaxen/CAGPJax/blob/dbddf3ac1d923e571bf542d3f95a5e34e3be87a2/src/cagpjax/linalg/eigh.py#L162. the gist: in a VJP, the derivative tells us how downstream code depended on the returned eigenvalues and eigenvectors. If it depended on it in a way that didn't depend on the degenerate eigenvectors themselves but rather just their subspace (which would include our case), then we can detect that and make the derivatives work numerically. If you like I can adapt that here. We would have two user configurable parameters: rtol for determining when an eigenvalue is close enough to 0 and grad_rtol for the equality check in the VJP.

The other challenge is working out the corresponding downdate formulas for eigh, but I think I have an idea how to do that.

sethaxen: I added a Nystrom implementation using eigh (not the one with the custom VJP). I locally ran some tests with low-rank operators, and it passed with flying colors. I'll clean these up a bit and commit them. For now, the Cholesky version is still there and selectable with a keyword argument, but if the eigh version really does pass all tests, then I think the Cholesky one should just be removed.

The downdate formulas ended up being pretty straightforward, but for completeness it would be good to drop the derivation somewhere. Is it best as a comment here?

pnkraemer: Really nice that `eigh` worked so well! And thanks for the pointers regarding the jax issue and your current version. 

Regarding the derivation, how about putting it as a comment in the src itself? That way it might age the best. Or is it too long for that?

> If you like I can adapt that here.

How about handling the pseudo-inverse with a callable function argument to `leave_one_out_xnystrace` instead? Currently, there is a keyword that calls functions with almost identical signatures. We could ask the user to pass one of these functions instead (with a suitable default), which would enable local, application-dependent workarounds like yours with `grad_rtol` defined in each application, rather than having the same workarounds in jax, cagp, and matfree. 
Something like: 
```python
def leave_one_out_xnystrace(..., nystrom: Callable[[Callable, Array, bool], tuple[Array, Array]]): ...

def nystrom_shift_and_cholesky(shift: float) -> Callable: ...

def nystrom_eigh() -> Callable: ...

# And then locally in user-code, until JAX merges the fix:
def nystrom_eigh_with_grad_rtol() -> Callable: ...
```

Matfree does something similar for matrix functions, where the "small dense matrix function" is an argument because sometimes, Pade approximations are better than SVDs and sometimes vice versa. Thoughts?

pnkraemer:
In general, how would you feel about merging the xtrace and xnystrace implementations once they converge, and handling the diagonal and rownorm estimators in separate PRs? This might simplify the review process and allow refactoring the existing MC-based estimators in the same module without worrying about annoying merge conflicts (which I currently worry about to be fully honest...). I also think that this stuff will be super useful even without diagonal estimation :)

matfree/stochtrace.py
sethaxen: > Regarding the derivation, how about putting it as a comment in the src itself? That way it might age the best. Or is it too long for that?

Sure, I'll give it a shot.

> How about handling the pseudo-inverse with a callable function argument to `leave_one_out_xnystrace` instead?

Ah, that's a nice idea. Sure, I'll do that.

sethaxen:
> In general, how would you feel about merging the xtrace and xnystrace implementations once they converge, and handling the diagonal and rownorm estimators in separate PRs?

I'm fine with it!

matfree/stochtrace.py
```diff
@@ -45,6 +45,406 @@ def estimate(matvecs, key, *parameters):
...
+        # If column k of Y is essential (i.e. rank(H)=rank(Hk)+1), then if we partition H as
+        # H=[Hk h; h' c], then its pseudoinverse is (Theorem 3; Albert, 1969)
+        # pinv(H) = [pinv(Hk) 0; 0 0] + avv',
+        # where v=pinv(H)[:, k] and a=c-h'pinv(H)h=1/pinv(H)[k,k] is the generalized Schur complement
+        # Then B_hat=Y pinv(H) Y' = Y_{-k} pinv(Hk) Y_{-k}' + a (Yv) (Yv)'
+        #                         = B_hat_{-k} + a (Yv) (Yv)'
+        # Let d=1 if column k is essential and 0 otherwise. Then
+        # B_hat_{-k} = B_hat - zz' where z=d Y pinv(H)[:, k]/sqrt(pinv(H)[k,k])
+        # If F and L are left-square root factors of B_hat and pinv(H), respectively, then
+        # z=d (F L[k, :]')/norm(L[k, :])
```
sethaxen: This derivation applies also for shifted Cholesky, so the API of the nystrom methods could be to return the left-square root factors of both the Nystrom approximation (F) and of pinv(H) (L), which I think would be a cleaner API. However, for shifted Cholesky, L is the inverse of the upper-triangular Cholesky factor U. It's generally more numerically stable to compute `F @ inv(U)` using a `solve` than inverting first. So I've opted for requiring the nystrom methods compute the downdate themselves and return it.

pnkraemer: Not sure I understand 100%, but the implementations look good to me 😅

matfree/stochtrace.py
```diff
@@ -45,6 +45,406 @@ def estimate(matvecs, key, *parameters):
...
+            return (
+                np.ones((num_samples,), dtype=B_mat.dtype) * linalg.trace(B_mat)
+            ).real
+
+        F, Z, correction = nystrom(matvec_flat, Omega)
+
+        if apply_resphering:
+            # Efficiently form T: the R factor of a QR decomposition of Omega
+            # Assumes Omega is well-conditioned (typically true for Gaussian/sphere samples if num_samples <= n/2)
+            T = linalg.cholesky(Omega.T.conj() @ Omega).T.conj()
```
pnkraemer: I think T could also be computed as 
```python
T = linalg.qr_r(Omega)

# Optional; if positive diagonals are needed:
signs = np.sign(linalg.diagonal(T))
T = T * np.where(signs == 0, 1, signs)[None, :]
```
because if we thin-QR decompose Omega = QR, then R^H Q^H Q R = R^H R, which implies R is exactly the T we're looking for. And the QR will always work even if Omega is ill-conditioned or we're in low precision. Would that work?

matfree/stochtrace.py
```diff
@@ -45,6 +45,406 @@ def estimate(matvecs, key, *parameters):
...
+        nystrom_left = linalg.solve_triangular(
+            chol_upper, Y_shifted.T.conj(), trans=2
+        ).T.conj()
+
+        Y_Hinv = linalg.solve_triangular(chol_upper, nystrom_left.T.conj()).T.conj()
+        Id = np.eye(chol_upper.shape[0], dtype=chol_upper.dtype)
+        chol_upper_inv = linalg.solve_triangular(chol_upper, Id)
+        downdate = (
+            Y_Hinv / func.vmap(linalg.vector_norm, in_axes=0)(chol_upper_inv)[None, :]
+        )
```
pnkraemer: Please don't judge, but I think this would look a lot nicer in two lines 😅 
```suggestion
        norms =  func.vmap(linalg.vector_norm, in_axes=0)(chol_upper_inv)
        downdate =  Y_Hinv / norms[None, :]
```
I typically find source code a lot easier to read without the autoformatter-induced linebreaks of long expressions with
```python
variable = (
    ...
)
```

But that's personal taste and doesn't affect the functionality, so feel free to ignore :)

matfree/stochtrace.py
```diff
@@ -45,6 +45,406 @@ def estimate(matvecs, key, *parameters):
...
+        # Let d=1 if column k is essential and 0 otherwise. Then
+        # B_hat_{-k} = B_hat - zz' where z=d Y pinv(H)[:, k]/sqrt(pinv(H)[k,k])
+        # If F and L are left-square root factors of B_hat and pinv(H), respectively, then
+        # z=d (F L[k, :]')/norm(L[k, :])
+        # Albert A. (1969), Conditions for Positive and Nonnegative Definiteness in Terms of Pseudoinverses.
+        # SIAM J. Appl. Math. 17(2), 434-440. doi:10.1137/0117041
+        downdate = (nystrom_left @ H_left_sqrt.T.conj()) / func.vmap(
+            linalg.vector_norm, in_axes=0
+        )(H_left_sqrt)
+        downdate = np.where(is_essential, downdate, 0.0)
```
pnkraemer: Like above
```suggestion
        # SIAM J. Appl. Math. 17(2), 434-440. doi:10.1137/0117041
         norms = func.vmap(linalg.vector_norm, in_axes=0)(H_left_sqrt)
        downdate = (nystrom_left @ H_left_sqrt.T.conj()) / norms
        downdate = np.where(is_essential, downdate, 0.0)
```

matfree/stochtrace.py
```diff
@@ -45,6 +45,406 @@ def estimate(matvecs, key, *parameters):
...
+            T = linalg.cholesky(Omega.T.conj() @ Omega).T.conj()
+            S = _qr_leave_one_out_factor(T)
+            # Omega projected onto the subspace spanned by Q_i, i.e. Q from qr(Omega_{-i}) leaving out Omega[:, i]
+            X = T - S * func.vmap(linalg.vdot, in_axes=1)(S, T)
+            # residual is B - B_hat_{-i}, where B_hat_{-i} approximates B leaving out Omega[:, i]
+            rank_residual = n - num_samples + 1
+            # squared norm of each sample after projection to the subspace spanned by the residual
+            sqnorm_samples_projected = np.sum(linalg.abs2(Omega), axis=0) - np.sum(
+                linalg.abs2(X), axis=0
+            )
```
pnkraemer: Like above (again, feel free to ignore)

```suggestion
            sqnorm_Omega = np.sum(linalg.abs2(Omega), axis=0)
            sqnorm_X = np.sum(linalg.abs2(X), axis=0)
            sqnorm_samples_projected = sqnorm_Omega - sqnorm_X 
```

matfree/stochtrace.py
sethaxen: That's right. The paper actually used QR, but in the thesis they switched to Cholesky decomposing the Gram matrix with this argument:
Screenshot_2026-06-04_07-57-41

Maybe on GPU the difference between QR and Cholesky isn't large enough to make a big difference, but I haven't checked. Also, the condition `num_samples > n/2` is one I haven't evaluated yet. While XNysTrace only requires `num_samples` matvecs, and so in principle it makes sense to not materialize the operator in this regime, I need to count up all of the other matrix operations; it might make sense, like XTrace, to materialize for `num_samples > n/2`. Combining that with the requirement that resphering is done with normal or sphere samples would mean that this Gram matrix would in practice always be well-conditioned. What do you think?

pnkraemer: I am not sure. In the codes I've worked with so far (mainly probabilistic numerics, PDE, or BNN stuff), I've never regretted switching from Cholesky to QR. While Cholesky is indeed faster, I have never found the runtime gains to be worth the increased numerical brittleness. While things might be different here because the matrices are well-conditioned, I would still lean towards QR over Cholesky, since we'll never have to worry about the stability of this line, which would improve maintainability.

We could also handle this case with a flag/enum, similar to the `implementation` argument in https://docs.jax.dev/en/latest/_autosummary/jax.lax.linalg.eig.html#jax.lax.linalg.eig. This would leave the speed/runtime to the user, and if we find that long term, one of them is never used, it can still be deleted. Thoughts?

sethaxen: > I would still lean towards QR over Cholesky, since we'll never have to worry about the stability of this line, which would improve maintainability.

Seems reasonable, with a comment explaining the choice



> We could also handle this case with a flag/enum, similar to the `implementation` argument in https://docs.jax.dev/en/latest/_autosummary/jax.lax.linalg.eig.html#jax.lax.linalg.eig. This would leave the speed/runtime to the user, and if we find that long term, one of them is never used, it can still be deleted. Thoughts?

Yeah, we could do that. I wonder though if a more matfree-style approach would be to accept a `qr_r` callable? Then we or someone else could implement a `qr_r_cholesky` that computes the R factor using a Cholesky decomposition.

> T = linalg.qr_r(Omega)

Just to clarify, you're suggesting `qr_r` be added to the `linalg` backend? In terms of implementation, would this just call `qr_reduced` internally and discard the `Q`, or would you call [`jax.numpy.linalg.qr`](https://docs.jax.dev/en/latest/_autosummary/jax.numpy.linalg.qr.html) with `mode='r'` to get just the `R`? This latter mode returns a rectangular matrix if the input matrix is rectangular, but we can always discard the extra rows/columns to make it like the reduced `R`. What do you think?

pnkraemer: > I wonder though if a more matfree-style approach would be to accept a qr_r callable?

:blush: 

Good call! I indeed prefer callables over flags, but this is just one line so I thought a flag might make sense. But I like the callable idea. 

We could accept a "triangularize: Callable", and default it to a QR decomposition, but if one wants cholesky, passing a `triangularize=lambda A: linalg.cholesky(A.T.conj() @ A)` shouldn't be very difficult. We might not even need to offer this cholesky function in the API; a sentence in the docstring could suffice.

> Just to clarify, you're suggesting qr_r be added to the linalg backend

I was convinced it was in there, but I now realise I might've thought about a different repo. (QR decompositions are quite common in probabilistic numerics, and I'm currently working more with probdiffeq than with matfree.) Sorry for the confusion. I think it would be useful to add it to the backend should we use it in stochtrace!

Here is a comparable backend with qr_r and also with a custom jvp that makes the code differentiable at 0: https://github.com/pnkraemer/probdiffeq/blob/main/probdiffeq/backend/linalg.py. This has worked well for a number of years in triangularization-heavy codes like square-root Kalman filters.

sethaxen: > We could accept a "triangularize: Callable", and default it to a QR decomposition, but if one wants cholesky, passing a `triangularize=lambda A: linalg.cholesky(A.T.conj() @ A)` shouldn't be very difficult. We might not even need to offer this cholesky function in the API; a sentence in the docstring could suffice.

So the thing is that what's actually happening in resphering is we are 1) (implicitly) projecting the test vectors onto the subspace spanned by the residual and 2) renormalizing them in this subspace. We don't *actually* do (1) because multiplying by the residual already does this, which is why we're able to get away with just computing a normalization term. However, the residual is the LOO residual, so we have `num_samples` different subspaces. To get the normalization term, we would need a bunch of different `Q_i` terms for the QR decomposition of `Omega_{-i}`, but we can downdate these from the full QR decomposition using the `R` term, and it turns out that we don't need the `Q` at all. So I wonder if `triangularize` is too vague to describe what this function must actually do. It specifically must compute the `R` term from some reduced QR decomposition. While we can document that `triangularize` must compute the `R` from a QR decomposition, I think it would be more literal to name it `qr_r`, especially if we have a `qr_r` function in the backend.

> Here is a comparable backend with qr_r and also with a custom jvp that makes the code differentiable at 0: https://github.com/pnkraemer/probdiffeq/blob/main/probdiffeq/backend/linalg.py.

Great! Here I don't think we would need the custom JVP, since the samples should never be 0 with non-zero probability. At least not for this application; maybe for if later features might call it on the origin, then it would be useful to add the custom rule then.

pnkraemer: Regarding the naming: yes, we can also call it qr_r. After all, this is what cholesky(Omega @ Omega.T) kind of does. And triangularisation might also be ambiguous (e.g. upper vs lower).

Regarding the JVP: I agree! No need for the custom JVP here.

sethaxen: I've added `qr_r` to the backend and a `qr_r` keyword.

matfree/stochtrace.py
```diff
@@ -45,6 +45,417 @@ def estimate(matvecs, key, *parameters):
...
+        Omega = func.vmap(lambda s: tree.ravel_pytree(s)[0])(samples).T
+        n, num_samples = Omega.shape
+
+        if num_samples > n:
+            raise ValueError(_error_num_samples(num_samples, maxval=n, minval=1))
+
+        def matvec_flat(v):
+            return tree.ravel_pytree(matvec(unflatten(v), *params))[0]
+
+        if num_samples == n:
```
sethaxen, line 264: Counting up all of the linear algrebraic operations that could end up as a fraction of cubic in `n` (as `num_samples` approaches `n`) and treating all of these as matmuls (i.e. having the same constant factors), we find that for `k=num_samples`, the estimate without resphering has about `n²k + 3nk² + k³`  FLOPS. The crossover point happens at `k>= (sqrt(2)-1)) *n`, i.e. `k >= 0.41*n`. Allocations are tricker to count up without knowing what XLA will do, but the cross-over point could be as low as `k >= 0.25*n`, so roughly, we could say:
- `k < 0.25*n`: estimate is more efficient and allocates less than direct
- `0.25*n <= k < 0.41*n`: estimate allocates more than direct but is more efficient
- `0.41 * n <= k <= n`: direct is more efficient and allocates less than estimate

Based on this, we could set the threshold to be `2.5 * num_samples >= n`, but since this is a bit heuristic, I'm wondering if it makes more sense to to just use `2 * num_samples >= n` like in the XTrace setting, which is a bit more conservative (in terms of preferring the estimating branch over the materialization branch).

sethaxen (commented):
Okay, I've added some comprehensive tests and incorporated the requested changes. Just about everything should be here now for this PR.

What I still plan to do:
- [ ] Nail down when we should switch to materialization in XNysTrace (see comment)
- [x] Split overly long lines into more shorter lines
- [x] Double-check all docs are up-to-date
- [x] Recreate the paper experiment figures with Float64 (we test single points on the plots, but the full visual check would be a good idea)

matfree/stochtrace.py
sethaxen: On second thought, a matvec cost could be much more expensive than a single matrix-vector product, so its probably best to take the branch that uses the smallest number of matvec evals, which will always be the estimate.

pnkraemer: Thanks for being so thorough! I wonder whether it's best to treat the number of matvecs as the runtime criterion, rather than the total number of matrix-matrix operations. In many cases, matvecs can be quite nasty (e.g., involving loops), and I found that to be the regular bottleneck. It also makes the math simpler, but that is only a side-effect :) 

The matvecs are the same across all xtrace/xnystrace algorithms, right? Y = A Omega. And the "large" allocations are always Y and Omega, and everything else is an operation on `num_samples` x `num_samples` matrices? If this is true, it might be best to have the same exact vs estimate decision as in xtrace, even if there are more of these O(num_samples^3) operations. Or am I oversimplifying this?

sethaxen: > The matvecs are the same across all xtrace/xnystrace algorithms, right?

No, XTrace requires `2*num_samples` matvecs (`Y = A @ Omega` and `Z = A @ Q`), while XNysTrace requires just `num_samples` (`Y = A @ Omega`). So for matvecs, XTrace has a cross-over at `2 * num_samples >= n`, while XNysTrace has it at `num_samples == n`. 

> And the "large" allocations are always Y and Omega, and everything else is an operation on `num_samples` x `num_samples` matrices?

For XTrace, `Omega`, `Y`, `Q`, and `Z` are `n x num_samples`. For XNysTrace, `Omega`, `Y`,`downdate`, `nystrom_left`, and (for shifted Cholesky), `Y_shifted`, and `Y_Hinv`. So if we're preferring estimate over direct, it makes sense to assume `nystrom_eigh`, which makes the cross-over for allocations the same as for XTrace `3 * num_samples >= n` (or `4 * num_samples >= n` if XLA is smart enough to compile away the allocation of `Omega` if we take the direct branch).

sethaxen:
Okay, I've done everything I've planned to, and barring the question of whether we should switch to the direct approach in XNysTrace for a smaller number of samples, the content of this PR should be complete.

Here are the experiment results, side-by-side with the corresponding figures (3 and 4) from the paper:
xtrace_experiments
Screenshot_2026-06-05_13-37-33
Screenshot_2026-06-05_13-37-51

The main difference I'm noticing is that for the "exp" experiment, our XNysTrace seems to plateau at ~10^-12 mean relative error instead of the ~10^-15 observed in the paper and achieved by XTrace. I suspect this is somehow due to the use of eigh (and the corresponding rtols) instead of the shifted Cholesky and will investigate this further.

sethaxen:
> The main difference I'm noticing is that for the "exp" experiment, our XNysTrace seems to plateau at ~10^-12 mean relative error instead of the ~10^-15 observed in the paper and achieved by XTrace. I suspect this is somehow due to the use of eigh (and the corresponding rtols) instead of the shifted Cholesky and will investigate this further.

Indeed, if we add XNysTrace with shifted Cholesky to the test, we see it behaves as in the paper experiments.
xtrace_experiment_exp

I tried different values of the `rtols` with `nystrom_eigh`, and they didn't have a significant effect on the resulting XNysTrace plateau. I did see that once we get to about `num_samples=80`, some of the eigenvalues of `H` drop below the threshold used to determine 0 eigenvalues, despite the fact that in exact math, the operator (and as a result `H`) is full-rank. So the issue is certainly floating-point error, but it's not obvious to me why the shifted version with Cholesky would behave better in this case.

At any rate, we do reproduce the paper's results when using the same method as the paper, and a mean relative error of 1e-12 is still extremely low, so I think our choice of defaulting to eigh is still the better one.

matfree/stochtrace.py
pnkraemer: Right, thanks for correcting me!

So should we keep the crossover at `num_samples == n` essentially due to the matvec argument? Or do you suggest something else?

pnkraemer:
These are really cool results! I agree that using eigh as a default is fine, even though its curve tapers off a bit before reaching machine precision. 10^-12 is small enough, and I like the default of not using Cholesky decompositions and not introducing shifts for numerical stability.

matfree/stochtrace.py
sethaxen: Yeah I think we should keep that. It's safest and closer to what the user has requested.

pnkraemer: Sounds good!

pnkraemer (approved):
This is really nice. Thank you for your work on this PR, and especially for taking the numerical details of this implementation so seriously!

I'll go ahead and merge it.

PR #264, opened 2026-05-19
feat(stochtrace): Add sampler for the unit sphere

sethaxen:
This PR adds `sampler_sphere` for generating samples on the scaled sphere $$\sqrt{n} \mathbb{S}^{n-1} \subset \mathbb{R}^n$$.

As indicated by https://www.ethanepperly.com/index.php/2024/01/28/dont-use-gaussians-in-stochastic-trace-estimation/, the variance of Hutchinson's estimator when using the sphere sampler is always lower than that when using the normal sampler. Since the estimator is unbiased, that means `sampler_sphere` will always produce a more accurate estimate on average than `sampler_normal` for the same number of matvecs.

### DIFF
--- a/matfree/stochtrace.py
+++ b/matfree/stochtrace.py
@@ -14,10 +14,7 @@ def estimator(integrand: Callable, /, sampler: Callable) -> Callable:
         [integrand_trace][matfree.stochtrace.integrand_trace].
         But any other integrand works, too.
     sampler
-        The sample function. Usually, either
-        [sampler_normal][matfree.stochtrace.sampler_normal],
-        [sampler_rademacher][matfree.stochtrace.sampler_rademacher],
-        or [sampler_sphere][matfree.stochtrace.sampler_sphere].
+        The sample function. See below for recommendations.
 
     Returns
     -------
@@ -26,8 +23,19 @@ def estimator(integrand: Callable, /, sampler: Callable) -> Callable:
         This function can be compiled, vectorised, differentiated,
         or looped over as the user desires.
 
-    """
+    Notes
+    -----
+    The statistical efficiency of the estimator for a given sampler depends on properties
+    of the operator, but we can provide some general advice. For an `n`-dimensional operator (see references):
+    - If the operator is real-valued and `n > O(100)`, use [sampler_rademacher][matfree.stochtrace.sampler_rademacher].
+    - If the operator is real-valued and `n < O(100)`, use [sampler_rademacher][matfree.stochtrace.sampler_rademacher] if the operator is known to be diagonal-dominant or [sampler_sphere][matfree.stochtrace.sampler_sphere] otherwise.
+    - If the operator is complex-valued, use [sampler_sphere][matfree.stochtrace.sampler_sphere] with a complex dtype.
 
+    References
+    ----------
+    - Epperly, E. (2023). [Stochastic trace estimation](https://www.ethanepperly.com/index.php/2023/01/26/stochastic-trace-estimation/).
+    - Epperly, E. (2024). [Don't use Gaussians in stochastic trace estimation](https://www.ethanepperly.com/index.php/2024/01/28/dont-use-gaussians-in-stochastic-trace-estimation/).
+    """
     def estimate(matvecs, key, *parameters):
         samples = sampler(key)
         Qs = func.vmap(lambda vec: integrand(matvecs, vec, *parameters))(samples)
--- a/matfree/stochtrace.py
+++ b/matfree/stochtrace.py
@@ -137,6 +137,11 @@ def sampler_rademacher(*args_like, num):
     return _sampler_from_jax_random(prng.rademacher, *args_like, num=num)
 
 
+def sampler_sphere(*args_like, num):
+    """Construct a function that samples from a unit sphere scaled to have identity covariance."""
+    return _sampler_from_jax_random(_random_sphere, *args_like, num=num)
+
+
 def _sampler_from_jax_random(sampler, *args_like, num):
     x_flat, unflatten = tree.ravel_pytree(*args_like)
 
@@ -145,3 +150,10 @@ def _sampler_from_jax_random(sampler, *args_like, num):
         return func.vmap(unflatten)(samples)
 
     return sample
+
+
+# sampler from the scaled sphere with the API of samplers in jax.random
+def _random_sphere(key, shape=(), dtype=None):
+    x = prng.normal(key, shape=shape, dtype=dtype)
+    sqrtn = np.sqrt(x.shape[-1])
+    return func.vmap(lambda v: v * (sqrtn / linalg.vector_norm(v)))(x)
--- a/matfree/stochtrace.py
+++ b/matfree/stochtrace.py
@@ -140,7 +140,17 @@ def sampler_rademacher(*args_like, num):
 
 def sampler_sphere(*args_like, num):
     """Construct a function that samples from a unit sphere scaled to have identity covariance."""
-    return _sampler_from_jax_random(_random_sphere, *args_like, num=num)
+    x_flat, unflatten = tree.ravel_pytree(*args_like)
+    n = x_flat.shape[0]
+    sqrtn = np.sqrt(n)
+
+    def sample(key):
+        samples = prng.normal(key, shape=(num, n), dtype=x_flat.dtype)
+        return func.vmap(lambda x: unflatten(x * (sqrtn / linalg.vector_norm(x))))(
+            samples
+        )
+
+    return sample
 
 
 def _sampler_from_jax_random(sampler, *args_like, num):
@@ -151,10 +161,3 @@ def _sampler_from_jax_random(sampler, *args_like, num):
         return func.vmap(unflatten)(samples)
 
     return sample
-
-
-def _random_sphere(key, shape=(), dtype=None):
-    """Sample from a scaled sphere with the API of the samplers in jax.random."""
-    x = prng.normal(key, shape=shape, dtype=dtype)
-    sqrtn = np.sqrt(x.shape[-1])
-    return func.vmap(lambda v: v * (sqrtn / linalg.vector_norm(v)))(x)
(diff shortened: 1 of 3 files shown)


sethaxen:
> `sampler_sphere` will always produce a more accurate estimate on average than `sampler_normal` for the same number of matvecs.

Would it be sensible then to update all tutorials that use `sampler_normal` specifically for trace estimation (the case explored in Epperly's post) to instead use `sampler_sphere`?

pnkraemer:
That would be great! 

Should we link Epperly's recommendations of using Rademacher or spherical samples somewhere? Eg in the existing estimator docstring?

matfree/stochtrace.py
```diff
@@ -145,3 +151,10 @@ def sample(key):
         return func.vmap(unflatten)(samples)
 
     return sample
+
+
+# sampler from the scaled sphere with the API of samplers in jax.random
+def _random_sphere(key, shape=(), dtype=None):
+    x = prng.normal(key, shape=shape, dtype=dtype)
+    sqrtn = np.sqrt(x.shape[-1])
+    return func.vmap(lambda v: v * (sqrtn / linalg.vector_norm(v)))(x)
```
pnkraemer: This implementation won't work unless `shape = (num, *x_flat_shape)` holds. Right? That means that, for its application, everything is good, but I wonder whether we should assert len(shape) == 2 or something similar somewhere (or leave a comment). to make it more obvious what this function assumes. Thoughts?

matfree/stochtrace.py
```diff
@@ -145,3 +151,10 @@ def sample(key):
         return func.vmap(unflatten)(samples)
 
     return sample
+
+
+# sampler from the scaled sphere with the API of samplers in jax.random
+def _random_sphere(key, shape=(), dtype=None):
+    x = prng.normal(key, shape=shape, dtype=dtype)
```
pnkraemer: I think this line could be a docstring. Something like this:
```suggestion
def _random_sphere(key, shape=(), dtype=None):
    """Sample from a scaled sphere with the API of the samplers in jax.random."""
    x = prng.normal(key, shape=shape, dtype=dtype)
```

tests/test_stochtrace/test_samplers.py
```diff
@@ -0,0 +1,38 @@
...
+from matfree import stochtrace
+from matfree.backend import np, prng, testing
+
+
+@testing.parametrize("n", [5])
+@testing.parametrize("dtype", [float, complex])
+def test_sampler_sphere(n, dtype):
+    """Assert that the sampler_sphere samples from a unit sphere scaled to have identity covariance."""
+
+    num_samples = 100000
```
pnkraemer: At first I thought "that's a lot of samples!" but then I checked and Matfree always uses that many for the trace-estimation tests... so all good here!

Minor suggestion for readability:
```suggestion
    num_samples = 100_000
```

tests/test_stochtrace/test_samplers.py
```diff
@@ -0,0 +1,38 @@
...
+        # hack to check that the result is complex
+        assert not np.allclose(x.imag, 0)
+    x2 = sampler(key1)
+    y = sampler(key2)
+    assert np.allclose(x, x2)
+    assert not np.allclose(x, y)
+    assert np.allclose(np.mean(x, axis=0), 0, atol=1e-2)
+    assert np.allclose(
+        (x.T.conj() @ x) / num_samples, np.eye(n).astype(dtype), atol=1e-2
+    )
```
pnkraemer: - How about reordering and documenting these tests a little bit? For example, the below would test samples the moment they are constructed (ie x first, then x2, then y). 

- Also, how about we use np.cov(x, axis=0) instead of computing the sample second moment manually? That might be easier to read, especially in relation to the first line. (Would require adding cov to backend; also note the manual dtype in eye(), which would also have to be added to the backend. But both I think would end up there eventually anyway)

For example:
```suggestion
    # Verify moments
    assert np.allclose(np.mean(x, axis=0), 0, atol=1e-2)
    assert np.allclose(np.cov(x, axis=0), np.eye(n, dtype=dtype), atol=1e-2)
    
    # Verify determinism (same key)
    x_again = sampler(key1)
    assert np.allclose(x, x_again)
    
    # Verify that samples differ if the key changes
    y = sampler(key2)
    assert not np.allclose(x, y)

```

pnkraemer (commented):
Thanks! Overall, this looks good to me. I left some minor suggestions, mainly regarding documentation. But I think the modifications should be relatively straightforward to implement.

matfree/stochtrace.py
sethaxen: So unlike the normal and rademacher cases, the sphere sampler is multivariate, so this follows the convention of e.g. https://docs.jax.dev/en/latest/_autosummary/jax.random.multivariate_normal.html and https://docs.jax.dev/en/latest/_autosummary/jax.random.dirichlet.html that the last dimension is the length of a single draw from the distribution. But maybe it makes more sense to follow the convention of https://docs.jax.dev/en/latest/_autosummary/jax.random.multinomial.html#jax.random.multinomial and have n be a required positional arg of the sampler. This is I think cleaner but then we can't use the wrapper `_sampler_from_jax_random`, and if not, then there's no benefit to having `_random_sphere` vs pushing this into `sampler_sphere` directly. What do you think?

pnkraemer: That makes sense. I'd say the simplest solution would be to just put it straight into sampler_sphere (as in, not using _sampler_from_jax_random)? Thanks for being so thorough with your replies!

sethaxen:
> Should we link Epperly's recommendations of using Rademacher or spherical samples somewhere? Eg in the existing estimator docstring?

Makes sense!

I was looking closer at the expressions Epperly gives for the (real) variances, and we additionally have

```math
\mathrm{Var}_{\text{rad}} < \mathrm{Var}_{\text{sphere}} \iff \sum_{i \ne j} A_{i j}^2 < \frac{n}{n+2} \left(\sum_{i \ne j} A_{i j}^2 + n \mathrm{Var}(\mathrm{diag}(A))\right)
```
As $`n \to \infty`$, $`n / (n+2) \to 1`$, and it becomes impossible for the sphere to have lower variance than Rademacher, while they'll have about the same variance whenever the diagonal of $`A`$ is constant. However, for moderate $`n`$, the sphere can indeed have lower variance but in the absolutely best case, just by a factor of $`n/(n+2)`$, which is quite modest. Personally, I'd then probably prefer `sampler_sphere` for quite small $n$ (say, $`n < O(100)`$) and otherwise `sampler_rademacher` (also avoids the issue of potentially redundant columns with Rademacher samples).

Additionally there's the generic advice that for complex operators, using complex test vectors reduces the variance by about half and should be preferred over real test vectors. (although currently `integrand_trace` and `integrand_diagonal` are broken for complex operators because they don't conjugate).

Would you be fine with me summarizing this advice in the docstring?

matfree/stochtrace.py
sethaxen: Sounds good!

> Thanks for being so thorough with your replies!

Of course! And likewise!

pnkraemer:
Yes, go for it!

Should we also open an issue for trace and diagonal being broken for complex values?

sethaxen:
Sure I'm happy to do so.

sethaxen:
I wanted to check whether the general advice for the Hutchinson estimator of the trace carried over to the estimator of the diagonal so that we could just make a blanket recommendation without limiting it to the trace. Accordingly, I worked out each of the variances. Given a diagonal estimate $`\hat{d}`$ for Hermitian $`A`$ (the only case we really need to consider, see Epperly's post) computed with $`k`$ test vectors, and focusing on the $`r`$th element, the estimator variances are:

| Test vectors | $`k \mathrm{Var}[\hat d_r]`$ |
|---|---:|
| Rademacher/Steinhaus | $`\sum_{i \ne r} \lvert A_{ri}\rvert^2`$ |
| Real Normal | $`\sum_{i \ne r} \lvert A_{ri}\rvert^2 + 2 A_{rr}^2`$ |
| Real Sphere | $`\frac{n}{n+2}\left(\sum_{i \ne r} \lvert A_{ri}\rvert^2 + 2\frac{n-1}{n} A_{rr}^2\right)`$ |
| Complex Normal | $`\sum_{i \ne r} \lvert A_{ri}\rvert^2 + A_{rr}^2`$ |
| Complex Sphere | $`\frac{n}{n+1}\left(\sum_{i \ne r} \lvert A_{ri}\rvert^2 + \frac{n-1}{n} A_{rr}^2\right)`$ |

So we find that the variances depend on the corresponding squared row/column norms and the squared diagonal. Additionally, the general properties and ranking of these variances is the same as for the trace (e.g. Rademacher's variance depends only on off-diagonal entries). Since the squared Frobenius norm estimator uses the trace estimator internally, this allows us to put in the `estimator` docstring general advice that applies to all of the currently supported integrands.

pnkraemer:
Wow, that's really nice! 

Btw, the tests failed for reasons outside of this PR. I pushed a fix in #265 and now things should be ok again.

matfree/stochtrace.py
```diff
@@ -25,8 +23,19 @@ def estimator(integrand: Callable, /, sampler: Callable) -> Callable:
...
         or looped over as the user desires.
 
-    """
+    Notes
+    -----
+    The statistical efficiency of the estimator for a given sampler depends on properties
+    of the operator, but we can provide some general advice. For an `n`-dimensional operator (see references):
+    - If the operator is real-valued and `n > O(100)`, use [sampler_rademacher][matfree.stochtrace.sampler_rademacher].
+    - If the operator is real-valued and `n < O(100)`, use [sampler_rademacher][matfree.stochtrace.sampler_rademacher] if the operator is known to be diagonal-dominant or [sampler_sphere][matfree.stochtrace.sampler_sphere] otherwise.
+    - If the operator is complex-valued, use [sampler_sphere][matfree.stochtrace.sampler_sphere] with a complex dtype.
```
sethaxen, line 32: This presumes complex operators are supported (i.e. #266 is fixed) and would be updated if the Steinhaus sampler is added (#267).

sethaxen (commented):
I've documented our recommendations RE samplers.

 > Btw, the tests failed for reasons outside of this PR. I pushed a fix in #265 and now things should be ok again.

Thanks! Locally the tests all pass for me now.

pnkraemer (approved):
Looks good, thanks! I'll merge once the CI passes.

PR #270, opened 2026-05-21
feat(stochtrace)!: Replace sampler_rademacher with sampler_signs

### LINKED ISSUE #267: feat(stochtrace): Add Steinhaus sampler
I propose we add `sampler_steinhaus` for the (complex) Steinhaus distribution (see https://www.ethanepperly.com/index.php/2023/01/26/stochastic-trace-estimation/).

The Steinhaus distribution is the complex generalization of the Rademacher distribution, being the uniform distribution on the complex unit circle (Rademacher is uniform on the 0-dimensional unit circle {-1,1}). As discussed in #264 and references linked therein, for a complex operator, Steinhaus will generally result in estimates having half the variance as using Rademacher. The estimator variance is always less than using the complex normal random sampler and is often less than using the complex sphere sampler (#264).

An alternative API would be to unify `sampler_rademacher` and `sampler_steinhaus` into `sampler_signs` and base the result on whether the dtype is real or complex.

sethaxen:
This is a proposed fix to #267 that adds a `sampler_signs` that is the same as the current `sampler_rademacher` for real dtypes and for complex dtypes samples from the Steinhaus distribution. This is breaking because it removes `sampler_rademacher`, but alternatively we could just deprecate `sampler_rademacher`.

Closes #267

### DIFF
--- a/matfree/stochtrace.py
+++ b/matfree/stochtrace.py
@@ -147,6 +147,14 @@ def sampler_rademacher(*args_like, num):
     return _sampler_from_jax_random(prng.rademacher, *args_like, num=num)
 
 
+def sampler_signs(*args_like, num):
+    """Construct a function that samples signs uniformly.
+
+    For real dtypes, this samples from a Rademacher distribution (uniformly over `{-1, 1}`). For complex dtypes, this samples from a Steinhaus distribution on the complex unit circle.
+    """
+    return _sampler_from_jax_random(_uniform_signs, *args_like, num=num)
+
+
 def sampler_sphere(*args_like, num):
     """Construct a function that samples from a unit sphere scaled to have identity covariance."""
     x_flat, unflatten = tree.ravel_pytree(*args_like)
@@ -172,3 +180,14 @@ def _sampler_from_jax_random(sampler, *args_like, num):
         return func.vmap(unflatten)(samples)
 
     return sample
+
+
+def _steinhaus(key, shape=(), dtype=None):
+    """Sample from a Steinhaus distribution on the complex unit circle."""
+    return np.sign(prng.normal(key, shape=shape, dtype=dtype))
+
+
+def _uniform_signs(key, shape=(), dtype=None):
+    if dtype is not None and np.dtype(dtype).kind == "c":
+        return _steinhaus(key, shape=shape, dtype=dtype)
+    return prng.rademacher(key, shape=shape, dtype=dtype)
--- a/matfree/stochtrace.py
+++ b/matfree/stochtrace.py
@@ -177,14 +177,14 @@ def _sampler_from_jax_random(sampler, *args_like, num):
     return sample
 
 
-def _steinhaus(key, shape=(), dtype=None):
+def _steinhaus(key, /, shape, dtype):
     """Sample from a Steinhaus distribution on the complex unit circle."""
     rdtype = np.dtype(dtype).type(0).real.dtype
     angle = prng.uniform(key, shape=shape, dtype=rdtype) * (2 * np.pi())
     return np.cos(angle) + 1j * np.sin(angle)
 
 
-def _uniform_signs(key, shape=(), dtype=None):
-    if dtype is not None and np.dtype(dtype).kind == "c":
+def _uniform_signs(key, /, shape, dtype):
+    if np.dtype(dtype).kind == "c":
         return _steinhaus(key, shape=shape, dtype=dtype)
     return prng.rademacher(key, shape=shape, dtype=dtype)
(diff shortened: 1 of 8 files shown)


matfree/stochtrace.py
```diff
@@ -172,3 +175,14 @@ def sample(key):
         return func.vmap(unflatten)(samples)
 
     return sample
+
+
+def _steinhaus(key, shape=(), dtype=None):
```
sethaxen: Alternatively this could be defined in the `prng` backend.

tests/test_stochtrace/test_samplers.py
```diff
@@ -41,3 +41,37 @@ def test_sampler_sphere_pytrees():
...
+        assert not np.allclose(x.imag, 0)
+    else:
+        assert x.dtype.kind != "c"
+        assert np.allclose(x.imag, 0)
+
+    # Verify moments
+    assert np.allclose(np.mean(x), 0, atol=1e-2)
+    assert np.allclose(
+        np.cov(x, rowvar=False, ddof=0), np.eye(n, dtype=dtype), atol=1e-2
+    )
```
sethaxen, line 69: These tolerances are arbitrary but work. Alternatively, we could pick atols using confidence intervals for Rademacher and uniformity tests on the unit circle for Steinhaus.

matfree/stochtrace.py
```diff
@@ -172,3 +175,14 @@ def sample(key):
...
+    """Sample from a Steinhaus distribution on the complex unit circle."""
+    return np.sign(prng.normal(key, shape=shape, dtype=dtype))
+
+
+def _uniform_signs(key, shape=(), dtype=None):
+    if dtype is not None and np.dtype(dtype).kind == "c":
+        return _steinhaus(key, shape=shape, dtype=dtype)
+    return prng.rademacher(key, shape=shape, dtype=dtype)
```
pnkraemer: Nice! Two quick questions:

1. Is there a specific reason why `_steinhaus` has default arguments?
2.  The `_steinhaus` code for real dtypes should be equivalent to the `prng.rademacher` code, right? If so, is there a specific reason for the if-clause, as opposed to always using the `_steinhaus` code?

tests/test_stochtrace/test_samplers.py
pnkraemer: Thanks for explaining! I'm okay with arbitrary tolerances. No need to increase sophistication without a concrete reason as far as I'm concerned :)

matfree/stochtrace.py
sethaxen: 1. To mirror the API of the samplers in jax.random, which have the same defaults. But I see that in the prng module the samplers have no default value for shape.
2. Yes, mathematically at least. But the Jax implementation of Rademacher uses Bernoulli sampling internally followed by an affine transform. I suspect this is more efficient, but more importantly, the result would be 0 with 0 probably even in floating point. On that note, maybe even for Steinhaus there should be a branch to replace zeros with 1 just in case. I haven't checked what the actual probability of getting a 0 would be with 32-bit float complex, but I'm guessing it's negligible.

pnkraemer: Thanks for the reply! Regarding the first point, I think that's valid. I did ask because the function is some hidden helper function that is called exactly once, so I am not sure it needs defaults. But I am okay with either solution.

Regarding the second point, I agree with you that even if there is a performance difference, it shouldn't matter much. However, after thinking about it some more, I believe it would make sense to replace zeros with ones just to be safe:

image

What do you think?

sethaxen: Agreed, I'll add that.

For Steinhaus, it might make sense to uniformly generate the angle and then either call sin and cos for the real and imaginary parts or exponentiate the pure imaginary. I'm not certain how performance and numeric stability of this would compare to generating two normals, so I'll run a benchmark on both CPU and GPU with both float32 and float64.

sethaxen: I ran a quick benchmark and found that the `exp` and `sincos` strategies are both clearly better than normalizing a normal and replacing 0 with 1. However, `exp` seems to be the best on CPU, while `sincos` is the clear winner on my GPU. My feeling is that `exp` and `sincos` are close enough on CPU that I'd prefer `sincos` simply because of this clear win on GPU.

**CPU**

benchmark_steinhaus

**GPU (GeForce RTX 3060)**

benchmark_steinhaus_gpu

pnkraemer: Thanks for the benchmarK! I am not sure I would describe the results as "clearly better" (except for double precision on GPUs, which is pretty rare as far as I know, but I might be wrong). Bu

t since they all look very similar to me, with an edge to sincos in the f64-gpu setting, I'd agree that this seems to be the best implementation choice :)

sethaxen: I switched to the sincos implementation and removed default keyword values from the internal methods.

pnkraemer (approved):
Thanks for taking care of that. I like the new API!

PR #279, opened 2026-06-09
Remove polyfill.js from extra_javascript

Ceyron:
The [website polyfill.io has been compromised](https://www.bleepingcomputer.com/news/security/polyfillio-javascript-supply-chain-attack-impacts-over-100k-sites/) and its JS delivery is injecting malicious code login popups into websites
image

It should be fine to remove it since it just ensures backward compatibility on some JS features which most modern browsers support by now.


pnkraemer (approved):
Oh wow, I had no idea. Thanks for catching that! I'll make a new release asap.

PR #280, opened 2026-06-09
feat(stochtrace): Add LOO-based estimators of the diagonal

### LINKED ISSUE #275: Implement the XDiag and XNysDiag estimators
Following #263, it would be useful to include estimators for `XDiag` and `XNysDiag` described in https://arxiv.org/abs/2301.07825 and https://arxiv.org/abs/2512.15929

Concretely, the result would be 2 functions meant to be passed to `estimator_leave_one_out`:
- `leave_one_out_xdiag`
- `leave_one_out_xnysdiag`

Both of these share a lot of code with the `XTrace` and `XNysTrace` algorithms, respectively. The sum of the estimate produced by `XDiag` is already an unbiased estimator of the trace but by the theory will have higher variance than `XTrace` (nevertheless, if one needs estimates of both the diagonal and the trace, taking this sum is probably a better use of matvecs). On the other hand, the sum of `XNysDiag`'s estimate is identical to the estimate produced by `XNysTrace`, so implementing these could involve a refactor of `XTrace` and `XNysTrace`.

One could also include `leave_one_out_xdiag_and_trace` (note: not xtrace) and `leave_one_out_xnysdiag_and_xnystrace` for convenience, but I wonder if it's better to include a tutorial that shows how the user can create their own `leave_one_out_` estimand that jointly estimates both.

sethaxen:
Following after #263, this PR adds
- [x] `leave_one_out_xdiag`
- [x] `leave_one_out_xnysdiag`

The implementations are pretty similar to those of `leave_one_out_xtrace` and `leave_one_out_xnystrace`, with a notable exception being that while the trace is always a scalar, the diagonal can be a pytree. To make this work, this PR also updates `estimator_leave_one_out` and `estimator_leave_one_out_mean_and_sem` to support pytrees.

EDIT: Also refactors tests for `leave_one_out_xtrace` and `leave_one_out_xnystrace` to reduce code duplication.

Closes #275

## Empirical comparison with thesis experiments

XDiag is introduced in the same paper as XTrace, while XNysDiag is introduced in Chapter 16 of Ethan Epperley's thesis. Figure 16.1 (below) shows experimental results, which I was able to closely replicate with the implementations in this PR:
Screenshot_2026-06-09_16-06-40 xdiag_experiment

EDIT: Like we saw with XNysTrace in https://github.com/pnkraemer/matfree/pull/263#issuecomment-4634481755, for the "exp" experiment, XNysDiag using Nystrom with eigh plateaus at a slightly higher abs relative error than using Nystrom with shifted Cholesky.


sethaxen:
The thesis does not discuss resphering in the context of the diagonal estimators, but it can be applied. For XDiag, it would require additional matvecs, so I don't think this makes sense to offer. I'm going to quickly look into whether it makes sense for XNysDiag.

tests/test_stochtrace/test_leave_one_out/test_estimator_loo.py
```diff
@@ -0,0 +1,44 @@
+"""Test estimator_leave_one_out and estimator_leave_one_out_mean_and_sem."""
+
+from matfree import stochtrace
+from matfree.backend import np, prng
+
+
+def test_estimator_leave_one_out_pytree_output():
```
pnkraemer, line 7: Thanks for testing the pytree compatibility changes! 

One suggestion: How about we make the matvecs pytree-valued in all (or a subset of) xtrace, xdiag, etc. tests and remove all "test_support_pytrees" style tests for the LOO estimators? This way, we'd get the same assertion with fewer tests, which will help reduce the test runtime. Thoughts?

tests/test_stochtrace/test_leave_one_out/test_xnysdiag.py
```diff
@@ -0,0 +1,206 @@
...
+    rdtype = np.abs(dtype(0)).dtype
+    n = 1000
+    m = 50
+    num_rep = 10
+    key = prng.prng_key(4)
+    key_mat, key = prng.split(key)
+    U = linalg.qr_reduced(prng.normal(key_mat, shape=(n, n), dtype=dtype))[0]
+    large_eigenvalues = np.ones(m, dtype=rdtype)
+    small_eigenvalues = np.ones(n - m, dtype=rdtype) * 1e-3
+    d = np.concatenate([large_eigenvalues, small_eigenvalues])
```
pnkraemer: Please correct me if I'm wrong, but it seems to me that the LOO tests repeatedly construct matrices with spectral drops, spectral decays, etc.. Feel free to move some of that logic to matfree.test_util, so that the tests become a bit easier to read :) (Or, if the existing test utilities help constructing those matrices, maybe we should use them?)

tests/test_stochtrace/test_leave_one_out/test_xnysdiag.py
```diff
@@ -0,0 +1,206 @@
...
+        nystrom=nystrom, apply_resphering=False
+    )
+    estimate_trace = stochtrace.estimator_leave_one_out(integrand_trace, sampler)
+    trace_result = estimate_trace(matvec, key_est, A)
+
+    trace_via_diag = np.sum(diag_result)
+    test_util.assert_allclose(trace_via_diag, trace_result)
+
+
+def test_xnysdiag_pytrees_supported(nystrom):
```
pnkraemer: Same comment as in the LOO pytree support test (maybe we can make these tests redundant and thus reduce the test runtime)

tests/test_stochtrace/test_leave_one_out/test_estimator_loo.py
sethaxen: Sure, I can do that. It won't make a difference for the trace estimators, which will always return a scalar, so I'd just make the change for the diagonal estimators introduced here.

tests/test_stochtrace/test_leave_one_out/test_xnysdiag.py
sethaxen: > Please correct me if I'm wrong, but it seems to me that the LOO tests repeatedly construct matrices with spectral drops, spectral decays, etc..

Yes, specifically, the matrices used in the "step" and "exp" experiments from the paper/thesis.

> Feel free to move some of that logic to matfree.test_util, so that the tests become a bit easier to read :)

Sure, I could do that. I wonder though, given that these are specific to testing this LOO family of estimators and likely wouldn't be used elsewhere, would it make more sense to make them fixtures in a `tests/test_stochtrace/test_leave_one_out/utils.py` file?

> (Or, if the existing test utilities help constructing those matrices, maybe we should use them?)

Ah, I see now I could be using `symmetric_matrix_from_eigenvalues` if updated to produce Hermitian matrices, except this would produce the same eigenvectors for all sets of eigenvalues, and would it be better for comprehensive testing to instead allow random generation of the eigenvectors by providing a key?

pnkraemer: That's a good idea! One could add a key or seed as an argument to the symmetric_matrix_from_eigvals-style functions that randomly samples Q like you do here. Would that make the creation of the test matrices in the LOO tests easier?

tests/test_stochtrace/test_leave_one_out/test_estimator_loo.py
pnkraemer: Fair point about the trace, however, the xtrace tests also include a "ensure Pytrees are supported" test which would be covered by the same logic even if the trace is a float. Traces are kind of independent of the current PR, though, so no need to expand the scope here unnecessarily :)

tests/test_stochtrace/test_leave_one_out/test_xnysdiag.py
sethaxen: I think so. The main thing I need for the experimental reproductions is the ability to pair a set of eigenvalues with the required tolerance of the (mean|median)-relative-error. Since the tolerances would vary for each estimator, I think it would still be good to have a central place to put a fixture or function that produces the required eigenvalues, and then the experimental reproduction could probably be implemented in each file as a fixture that returns a tuple `(matrix, rtol)` using `symmetric_matrix_from_eigenvalues` internally. Should it be renamed `hermitian_matrix_from_eigenvalues`?

pnkraemer: > Should it be renamed hermitian_matrix_from_eigenvalues

Yes, I think so, especially now that support for complex numbers is growing in matfree!

> and then the experimental reproduction could probably be implemented in each file as a fixture that returns a tuple (matrix, rtol)

Sounds good! I like to use test-cases for that (`testing.parametrize_with_cases`, see e.g. the decomposition tests, which use this construction), but of course choose the tool you prefer here :)

sethaxen:
I spent some time looking into resphering for the diagonal estimators. The key observation is that
```math
\begin{aligned}
(B \omega) \circ \omega &= \mathrm{diag}(B \omega \omega^\dagger)\\
\mathrm{diag}(B \omega \omega^\dagger)_j &= e_j^\dagger B \omega \omega^\dagger e_j = \omega^\dagger (e_j e_j^\dagger B) \omega
\end{aligned}
```
so when using a Hutchinson-style estimator for the diagonal, what we're really doing is simultaneously estimating a bunch of traces for the operators $`C_j=E_j B`$, where $`E_j = e_j e_j^\top`$. We can use this result to work out the variances of the estimators for the various test vectors (and it's a lot easier than what I did to get https://github.com/pnkraemer/matfree/pull/264#issuecomment-4506386875).

When resphering, I think it's a mistake to try to project each test vector onto an orthoprojector $`P`$ such that $`BP=B`$ like we do for trace estimates. Rather, I suspect that for each diagonal entry, we want an orthoprojector such that $` P C_j = C_j P = C_j`$. If we have this, then Corollary 13.4 of the thesis would follow, and resphering would reduce the variance of the estimator. If we have an orthonormal matrix such that $`BQ=0`$, then it seems we could construct such a projector as
```math
P = I - Q (I - f f^\dagger) Q^\dagger, \qquad f = \frac{Q^\dagger e_j}{\lVert Q^\dagger e_j \rVert}
```
(Yeah this seems to just be another QR downdate formula).

What's tricky is that for each residual/test vector pair, we need to resphere the test vector differently for each diagonal entry.  It's quite possible this can be efficiently evaluated using just quantities we already have, at least for XNysDiag, but since this is going well beyond the source material, I'm going to shelf this idea for another time.

pnkraemer:
Thanks for checking! 

> but since this is going well beyond the source material, I'm going to shelf this idea for another time

I did get a similar feeling reading through the derivation/sketch. It's nice that it seems possible, but I agree with you, it should probably be handled separately (if at all).

sethaxen:
I refactored the new diagonal tests using `parametrize_with_cases`. Does the overall approach look like what you had in mind? If so, would you like me to also refactor the xtrace and xnystrace tests here?

I've opted to keep xdiag and xnysdiag tests in their own files, but some of the tests could in principle be merged if we lumped them into the same test file, but not all of them. What are your thoughts?

tests/test_stochtrace/test_leave_one_out/conftest.py
```diff
@@ -0,0 +1,25 @@
...
+def exp_eigvals(n, /):
+    """Eigenvalues that decay rapidly."""
+    return 0.7 ** np.arange(n)
+
+
+def step_eigvals(n, /, *, num_flat=50, drop_value=1e-3):
+    """Eigenvalues that are flat with a sudden drop."""
+    eigvals_flat = np.ones(num_flat)
+    eigvals_drop = np.ones(n - num_flat) * drop_value
+    return np.concatenate([eigvals_flat, eigvals_drop])
```
pnkraemer: Random thought: I think these could be useful across matfree - should we move these matrices to the test_utils and call the test_utils in the LOO experiments instead of importing from the conftest? 

For example, something like `hermitian_matrix_eigvals_decaying` and `hermitian_matrix_eigvals_step` with exactly these parameters here + the key that is then passed to the `hermitian_matrix_from_eigvals`? I think this might be easier to read than the conftest.

I can anticipate using them in the Lanczos tests as well in the future, where I think Matfree currently hacks something very similar together.

tests/test_stochtrace/test_leave_one_out/conftest.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+@testing.fixture(name="nystrom")
+@testing.parametrize(
+    "factory", [stochtrace.nystrom_eigh, stochtrace.nystrom_shifted_cholesky]
+)
+def fixture_nystrom(factory):
```
pnkraemer: I wonder whether a testing.parametrize(nystrom, [stochtrace.nystrom_eigh, stochtrace.nystrom_shifted_cholesky) placed on top of each xnysdiag and xnystrace test is more readable than this fixture that wraps parameters like this. What do you think?

tests/test_stochtrace/test_leave_one_out/test_xnysdiag.py
```diff
@@ -0,0 +1,197 @@
+"""Tests for leave_one_out_xnysdiag."""
+
+from matfree import stochtrace, test_util
+from matfree.backend import func, linalg, np, prng, testing
+
+from .conftest import exp_eigvals, step_eigvals
```
pnkraemer: If I remember correctly, one doesn't have to import from the conftest but anything in the conftest.py is automatically detected by pytest. (I might be wrong, but it could be worth checking.)

That said, depending on what comes out of the comments in the conftest file, this comment here might become irrelevant :)

pnkraemer:
Yes, I think it is nice and clear now. No need to merge, I am okay with xnysdiag and xdiag in separate files.

tests/test_stochtrace/test_leave_one_out/test_xnysdiag.py
sethaxen: One doesn't need to import fixtures (or maybe cases?), but in this case these are normal functions that we call within the cases, so we did need to import them. That being said, `conftest.py` is now removed, so we don't have these anymore.

tests/test_stochtrace/test_leave_one_out/conftest.py
sethaxen: Yeah that was pretty straightforward, so I switched to it.

tests/test_stochtrace/test_leave_one_out/conftest.py
sethaxen: Added! Does it look like what you were thinking?

sethaxen:
I implemented the requested changes to the test.

> would you like me to also refactor the xtrace and xnystrace tests here?

Would you like this here or in a separate PR?

matfree/test_util.py
```diff
@@ -3,23 +3,32 @@
...
+    if dtype is None:
+        dtype = eigvals.dtype
+    eigvals = eigvals.real
+    Q, _ = linalg.qr_reduced(prng.normal(key, shape=(n, n), dtype=dtype))
+    return (Q * eigvals) @ Q.T.conj()
+
+
+def hermitian_matrix_eigvals_decaying(n, /, key, *, dtype=None):
+    """Hermitian matrix whose eigenvalues decay geometrically (0.7^k)."""
+    eigvals = 0.7 ** np.arange(n)
```
pnkraemer: How about including 0.7 in the arguments? We can also do this in the future as soon as we need to change it, but while we're at it... 

```suggestion
def hermitian_matrix_eigvals_decaying(n, /, key, *, base=0.7, dtype=None):
    """Hermitian matrix whose eigenvalues decay geometrically (x^k)."""
    eigvals = x ** np.arange(n)
```

tests/test_stochtrace/test_leave_one_out/conftest.py
pnkraemer: Great!

pnkraemer:
Thanks! 

> Would you like this here or in a separate PR?

Both are fine. I'd say if you're doing another (few) commits for this PR, we can handle them in the same sweep. Otherwise, happy to review a separate one as well :)

sethaxen:
> Both are fine. I'd say if you're doing another (few) commits for this PR, we can handle them in the same sweep. Otherwise, happy to review a separate one as well :)

Okay, I refactored the x(nys)trace tests similarly to how I had refactored the corresponding diag tests. For all 4 sets of tests, I also adapted the corresponding exact-depending-on-num-samples test to test if we have a pytree with multiple leaves, so now all pytree-compatibility-specific tests have been removed.

pnkraemer (approved):
Awesome! Thanks a lot :)

PR #285, opened 2026-06-24
feat(stochtrace): Add estimators of the squared row norm

### LINKED ISSUE #284: Implementing row-norm-squared estimators
Along the lines of the existing `monte_carlo_frobeniusnorm_squared`, it would be useful to include a `monte_carlo_rownorms_squared` integrand to the `stochtrace` module, as well as the `XRowNorm`/`XSymRowNorm` estimators described in Chapter 17 of https://arxiv.org/abs/2512.15929.

For the LOO-based estimators, it may make sense to depart from our current convention of using the name given in the thesis. `XRowNorm` is somewhat misleading given that it estimates the *squared* row-norm. Additionally, as noted in Remark 17.4, `XSymRowNorm` works not just for symmetric/hermitian matrices but also for any normal or almost-Hermitian matrix. Additionally, `XRowNorm` and `XSymRowNorm` both are based on the randomized SVD and primarily differ in how `Q` is constructed, so I think the two could be unified into a single  integrand `leave_one_out_xrownorms_squared` with a flag `is_normal: bool=false`.

One could also add column norm squared estimators, but it's straightforward to document that the user can do this by passing the adjoint of their matvec to the estimator. One could alternatively implement column norms first and document that the user estimates row norms via the adjoint, but IMO the row norm estimators are more natural, since the resulting PyTree is the same as the output of the matvec, and the first step in the estimate is calling the matvec.

sethaxen:
As proposed in #284, this PR adds
- [x] `monte_carlo_rownorms_squared` (JL)
- [x] `leave_one_out_xrownorms_squared`
- [x] `leave_one_out_xsymrownorms_squared`

Closes #284 

## Empirical comparison with thesis experiments

The LOO estimator(s) are introduced in Chapter 17 of Ethan Epperley's thesis, and Figure 17.2 (below) compares their performance on the same test matrices we've been using in the other PRs. Below I ran the same experiments with these implementations and saw nearly identical results:
Screenshot_2026-06-24_18-51-05
fig17-2_repro

### DIFF
--- a/matfree/backend/func.py
+++ b/matfree/backend/func.py
@@ -46,6 +46,18 @@ def linear_transpose(func, *primals):
     return jax.linear_transpose(func, *primals)
 
 
+def linear_adjoint(func, *primals):
+    transpose = jax.linear_transpose(func, *primals)
+
+    def tree_conj(x):
+        return jax.tree.map(lambda x: x.conj(), x)
+
+    def adjoint(*primal_results):
+        return tree_conj(transpose(*tree_conj(primal_results)))
+
+    return adjoint
+
+
 def grad(func):
     return jax.grad(func)
 
--- a/matfree/stochtrace.py
+++ b/matfree/stochtrace.py
@@ -530,17 +530,8 @@ def leave_one_out_xnysdiag(
     return integrand
 
 
-def leave_one_out_xrownorms_squared(*, is_normal: bool = False) -> Callable:
-    """Construct an integrand for estimating squared row norms using the XRowNorm/XSymRowNorm algorithms (Epperly, 2025).
-
-    Parameters
-    ----------
-    is_normal
-        `True` indicates that the operator is normal (or almost-Hermitian), which allows for the more sample-efficient XSymRowNorm construction.
-        `False` uses the more general XRowNorm construction.
-
-        An operator `matvec` is normal if it commutes with its adjoint (i.e. conjugate-transpose-conjugate) operation `matvec_adj`.
-        That is, applying `matvec` followed by `matvec_adj` is equivalent to applying `matvec_adj` followed by `matvec`, up to numerical precision.
+def leave_one_out_xrownorms_squared() -> Callable:
+    """Construct an integrand for estimating squared row norms using the XRowNorm algorithm (Epperly, 2025).
 
     Returns
     -------
@@ -555,12 +546,48 @@ def leave_one_out_xrownorms_squared(*, is_normal: bool = False) -> Callable:
     -----
     To estimate squared column norms instead, pass the adjoint (i.e. conjugate-transpose-conjugate) of the matvec.
 
+    For normal operators (those that commute with their adjoint),
+    [leave_one_out_xsymrownorms_squared][matfree.stochtrace.leave_one_out_xsymrownorms_squared],
+    requires fewer matvecs per sample but is typically less accurate per sample.
+
     References
     ----------
     - Epperly EN (2025). Make the most of what you have: Resource-efficient randomized algorithms for matrix computations. PhD Thesis.
         arXiv: [2512.15929](https://arxiv.org/abs/2512.15929)
     """
+    return _leave_one_out_rownorms_squared(iterate_subspace=True)
 
+
+def leave_one_out_xsymrownorms_squared() -> Callable:
+    """Construct an integrand for estimating squared row norms using the XSymRowNorm algorithm (Epperly, 2025).
+
+    Returns
+    -------
+    integrand
+        An integrand function compatible with
+        [estimator_leave_one_out][matfree.stochtrace.estimator_leave_one_out]
+        whose input has the signature ``(matvec, samples, *params)`` and whose output is a
+        pytree with each leaf having shape ``(num_samples, n_k)``, giving one squared-row-norm
+        estimate per leave-one-out sample.
+        The ``matvec`` must be a normal operator, i.e. it must commute with its adjoint
+        (conjugate-transpose-conjugate) up to numerical precision.
+
+    Notes
+    -----
+    To estimate squared column norms instead, pass the adjoint (i.e. conjugate-transpose-conjugate) of the matvec.
+
+    For general (non-normal) operators, use
+    [leave_one_out_xrownorms_squared][matfree.stochtrace.leave_one_out_xrownorms_squared].
+
+    References
+    ----------
+    - Epperly EN (2025). Make the most of what you have: Resource-efficient randomized algorithms for matrix computations. PhD Thesis.
+        arXiv: [2512.15929](https://arxiv.org/abs/2512.15929)
+    """
+    return _leave_one_out_rownorms_squared(iterate_subspace=False)
+
+
+def _leave_one_out_rownorms_squared(*, iterate_subspace: bool):
     def integrand(matvec, samples, *params):
         sample0 = tree.tree_map(lambda s: s[0], samples)
         _, unflatten = tree.ravel_pytree(sample0)
@@ -578,9 +605,13 @@ def leave_one_out_xrownorms_squared(*, is_normal: bool = False) -> Callable:
         def matvec_flat(v):
             return tree.ravel_pytree(matvec(unflatten(v), *params))[0]
 
-        num_matvecs = (2 if is_normal else 3) * num_samples
+        def matvec_flat_adjoint(v):
+            return tree.ravel_pytree(func.linear_adjoint(matvec_flat, v))[0]
 
-        if num_matvecs >= n:
+        matmat = func.vmap(matvec_flat, in_axes=-1, out_axes=-1)
+
+        num_matvecs_per_sample = 2 + int(iterate_subspace)
+        if num_matvecs_per_sample * num_samples >= n:
             # It's faster, more accurate, and allocates less memory to compute the row
             # norms exactly and deterministically on the materialized operator when
             # the number of matvecs exceeds the input dimension of the operator
@@ -592,14 +623,11 @@ def leave_one_out_xrownorms_squared(*, is_normal: bool = False) -> Callable:
             )
             return func.vmap(unflatten_out)(srn_all.T)
 
-        matmat = func.vmap(matvec_flat, in_axes=-1, out_axes=-1)
-        matvec_flat_adjoint = func.linear_adjoint(matvec_flat, Omega[:, 0])
-
         G = matmat(Omega)
-        if is_normal:
-            Y = G
+        if iterate_subspace:
+            Y = func.vmap(matvec_flat_adjoint, in_axes=-1, out_axes=-1)(G)
         else:
-            (Y,) = func.vmap(matvec_flat_adjoint, in_axes=-1, out_axes=-1)(G)
+            Y = G
         Q, R = linalg.qr_reduced(Y)
 
         Z = matmat(Q)
--- a/matfree/stochtrace.py
+++ b/matfree/stochtrace.py
@@ -544,11 +544,13 @@ def leave_one_out_xrownorms_squared() -> Callable:
 
     Notes
     -----
-    To estimate squared column norms instead, pass the adjoint (i.e. conjugate-transpose-conjugate) of the matvec.
+    To estimate squared column norms instead, pass the adjoint (i.e. conjugate-transpose-conjugate) of the matvec (
+    see [`func.linear_adjoint`][matfree.func.linear_adjoint]).
 
     For normal operators (those that commute with their adjoint),
     [leave_one_out_xsymrownorms_squared][matfree.stochtrace.leave_one_out_xsymrownorms_squared],
-    requires fewer matvecs per sample but is typically less accurate per sample.
+    performs only 2/3 of the matvecs per sample compared to this algorithm but is typically less
+    accurate per sample.
 
     References
     ----------
@@ -559,7 +561,7 @@ def leave_one_out_xrownorms_squared() -> Callable:
 
 
 def leave_one_out_xsymrownorms_squared() -> Callable:
-    """Construct an integrand for estimating squared row norms using the XSymRowNorm algorithm (Epperly, 2025).
+    """Construct an integrand for estimating squared row norms of a normal operator using the XSymRowNorm algorithm (Epperly, 2025).
 
     Returns
     -------
@@ -569,12 +571,13 @@ def leave_one_out_xsymrownorms_squared() -> Callable:
         whose input has the signature ``(matvec, samples, *params)`` and whose output is a
         pytree with each leaf having shape ``(num_samples, n_k)``, giving one squared-row-norm
         estimate per leave-one-out sample.
+        The input and output pytree structures are assumed to be identical.
         The ``matvec`` must be a normal operator, i.e. it must commute with its adjoint
         (conjugate-transpose-conjugate) up to numerical precision.
 
     Notes
     -----
-    To estimate squared column norms instead, pass the adjoint (i.e. conjugate-transpose-conjugate) of the matvec.
+    For normal operators, the row and column norms are equal, so this algorithm can also be used to estimate squared column norms.
 
     For general (non-normal) operators, use
     [leave_one_out_xrownorms_squared][matfree.stochtrace.leave_one_out_xrownorms_squared].
--- a/matfree/stochtrace.py
+++ b/matfree/stochtrace.py
@@ -319,15 +319,11 @@ def leave_one_out_xdiag() -> Callable:
                 np.ones((num_samples, 1), dtype=diag_B.dtype) * diag_B
             )
 
-        matvec_flat_transpose = func.linear_transpose(matvec_flat, Omega[:, 0])
-
-        def matvec_flat_adjoint(v):
-            (result,) = matvec_flat_transpose(v.conj())
-            return result.conj()
+        matvec_flat_adjoint = func.linear_adjoint(matvec_flat, Omega[:, 0])
 
         Y = func.vmap(matvec_flat, in_axes=-1, out_axes=-1)(Omega)
         Q, R = linalg.qr_reduced(Y)
-        Z = func.vmap(matvec_flat_adjoint, in_axes=-1, out_axes=-1)(Q)
+        (Z,) = func.vmap(matvec_flat_adjoint, in_axes=-1, out_axes=-1)(Q)
 
         def _diag_exact():
             diag_B = func.vmap(linalg.vdot, in_axes=0)(Z, Q)
@@ -597,18 +593,13 @@ def leave_one_out_xrownorms_squared(*, is_normal: bool = False) -> Callable:
             return func.vmap(unflatten_out)(srn_all.T)
 
         matmat = func.vmap(matvec_flat, in_axes=-1, out_axes=-1)
-
-        matvec_flat_transpose = func.linear_transpose(matvec_flat, Omega[:, 0])
-
-        def matvec_flat_adjoint(v):
-            (result,) = matvec_flat_transpose(v.conj())
-            return result.conj()
+        matvec_flat_adjoint = func.linear_adjoint(matvec_flat, Omega[:, 0])
 
         G = matmat(Omega)
         if is_normal:
             Y = G
         else:
-            Y = func.vmap(matvec_flat_adjoint, in_axes=-1, out_axes=-1)(G)
+            (Y,) = func.vmap(matvec_flat_adjoint, in_axes=-1, out_axes=-1)(G)
         Q, R = linalg.qr_reduced(Y)
 
         Z = matmat(Q)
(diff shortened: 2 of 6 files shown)


matfree/stochtrace.py
```diff
@@ -533,6 +534,116 @@ def matvec_flat(v):
...
+        arXiv: [2512.15929](https://arxiv.org/abs/2512.15929)
+    """
+
+    def integrand(matvec, samples, *params):
+        sample0 = tree.tree_map(lambda s: s[0], samples)
+        _, unflatten = tree.ravel_pytree(sample0)
+
+        out_shape = func.eval_shape(matvec, sample0, *params)
+        out0 = tree.tree_map(lambda s: np.zeros(s.shape, dtype=s.dtype), out_shape)
+        _, unflatten_out = tree.ravel_pytree(out0)
```
sethaxen, line 600: Maybe there's a less clunky way to do this?

matfree/stochtrace.py
```diff
@@ -533,6 +534,116 @@ def matvec_flat(v):
...
+
+            # srn(B_hat) leaving out one sample
+            srn_B_hat_loo = srn_B_hat[:, None] - linalg.abs2(Z @ S)
+            # Johnson-Lindenstrauss estimate of srn(B - B_hat_{-i}) using samples[i, :] as probes.
+            srn_residual_loo = linalg.abs2(G - Z @ X)
+            return srn_B_hat_loo + srn_residual_loo
+
+        is_Y_rank_deficient = (
+            np.sum(np.abs(linalg.diagonal(R)) > np.finfo_eps(R.dtype)) < num_samples
+        )
```
sethaxen, line 656: XTrace should actually get this treatment as well, since from the diagonal of `R`, we cannot determine the rank of `R`, just whether it is rank-deficient.

matfree/stochtrace.py
sethaxen: Note to self: xdiag and xnysdiag assume that the input and output pytrees are identical, but this is stricter than the actual requirement that they have the same number of leaves. That's probably a bug that should be fixed.

matfree/stochtrace.py
```diff
@@ -533,6 +534,116 @@ def matvec_flat(v):
...
+        _, unflatten = tree.ravel_pytree(sample0)
+
+        out_shape = func.eval_shape(matvec, sample0, *params)
+        out0 = tree.tree_map(lambda s: np.zeros(s.shape, dtype=s.dtype), out_shape)
+        _, unflatten_out = tree.ravel_pytree(out0)
+
+        Omega = func.vmap(lambda s: tree.ravel_pytree(s)[0])(samples).T
+        n, num_samples = Omega.shape
+
+        if num_samples > n:
```
sethaxen, line 605: A normal operator must have the same number of input and output leaves. If `is_normal=True` is set and this is not true, there will be an error, but maybe we should raise a more informative one ourselves? (technically all of the other LOO-based estimators could do the same).

matfree/stochtrace.py
```diff
@@ -533,6 +534,116 @@ def matvec_flat(v):
...
+            return func.vmap(unflatten_out)(srn_all.T)
+
+        matmat = func.vmap(matvec_flat, in_axes=-1, out_axes=-1)
+
+        matvec_flat_transpose = func.linear_transpose(matvec_flat, Omega[:, 0])
+
+        def matvec_flat_adjoint(v):
+            (result,) = matvec_flat_transpose(v.conj())
+            return result.conj()
+
```
sethaxen: I wonder if it would be useful to add to `func` a utility that gives the adjoint of a linear operator, handling the conjugating and such? We would use it several places here, and it might be useful for users who have complex operators.

matfree/stochtrace.py
sethaxen: Actually, I think for xdiag and xnysdiag, we have three choices:
- require that the input and output pytrees have the same structure
- choose that the diagonal has the structure of either the input or the output pytree (we currently give it the input structure)
- allow the user to choose using a flag

matfree/stochtrace.py
```diff
@@ -533,6 +534,116 @@ def matvec_flat(v):
...
 
 
+def leave_one_out_xrownorms_squared(*, is_normal: bool = False) -> Callable:
+    """Construct an integrand for estimating squared row norms using the XRowNorm/XSymRowNorm algorithms (Epperly, 2025).
+
+    Parameters
+    ----------
+    is_normal
+        `True` indicates that the operator is normal (or almost-Hermitian), which allows for the more sample-efficient XSymRowNorm construction.
+        `False` uses the more general XRowNorm construction.
```
pnkraemer: What's your take on splitting this function up into `leave_one_out_xrownorm_squared` and `leave_one_out_xsymrownorm_squared`? I understand that the implementations are similar, but both algorithms have different mathematical assumptions and thus, technically, compute different things. 

I wonder whether it would be easier to test and document the code if we didn't switch using a flag. But I am open to both. What do you think?

matfree/stochtrace.py
pnkraemer: I'm not sure I can follow. Are you discussing whether or not the "unflatten_out" can be skipped and "unflatten" can be used instead, in relation to mathematical assumptions about the linear operator? 

And why is the requirement only that the number of leaves is similar, not the full pytree structure?

matfree/stochtrace.py
pnkraemer: (Like in the comment above). Why only the same number of leaves? I would expect the pytree structures to be identical.

matfree/stochtrace.py
```diff
@@ -533,6 +534,116 @@ def matvec_flat(v):
...
+
+        if num_samples > n:
+            raise ValueError(_error_num_samples(num_samples, maxval=n, minval=1))
+
+        def matvec_flat(v):
+            return tree.ravel_pytree(matvec(unflatten(v), *params))[0]
+
+        num_matvecs = (2 if is_normal else 3) * num_samples
+
+        if num_matvecs >= n:
```
pnkraemer: We seem to do this for all LOO estimators (and should possibly do this for all Monte Carlo estimators too.). I wonder whether it's time to automate this corner case. For example, flattening Omega could happen in `estimator_leave_one_out` and we catch the `num_matvecs >= n` there? (This would require playing with the data structures, ie moving num_matvecs and _rownorms_squared to the integrand and would thus be a slightly larger change. We could resolve this separately from this PR.)

matfree/stochtrace.py
pnkraemer: Yes, I think that's a good idea. JAX's handling of complex numbers is a bit unfriendly for complex linear algebra, and wrapping it in a more meaningful API is one of the reasons I channel all imports through the backend anyway. 

So yes, how about we add a `linear_adjoint` to the backend that has the same API as `jax.linear_transpose` but handles conjugation?

matfree/stochtrace.py
pnkraemer: I wonder whether this is another thing to move out of the integrand definitions, since it repeats itself as well. (See comment above.)

But again, not sure this should be handled in this PR.

tests/test_stochtrace/test_leave_one_out/test_xrownorms_squared.py
```diff
@@ -0,0 +1,153 @@
...
+    """Hermitian matrix with eigenvalues that are flat with a sudden drop, XRowNorm"""
+    return test_util.hermitian_matrix_eigvals_step, 60, 1e-5, False
+
+
+@testing.parametrize("dtype", [float, complex])
+@testing.parametrize_with_cases(
+    "make_A, num_samples, max_rel_err, is_normal",
+    cases=".",
+    prefix="cases_experiments_",
+)
```
pnkraemer: Do you think it would be clearer if we moved the `dtype` to the case declarations? We can use `testing.parametrize` on the cases as well. 

There is also the option of making the test-case outputs into a dedicated datastructure (eg a dataclass) with those 5 fields, in which case we dont have to remember what the length-4/5 tuple stands for :)

tests/test_stochtrace/test_leave_one_out/test_xrownorms_squared.py
```diff
@@ -0,0 +1,153 @@
...
+def test_xrownorms_squared_reproduce_experiments(
+    make_A, num_samples, max_rel_err, is_normal, dtype
+):
+    """Assert that squared row norms are estimated accurately for structured spectra."""
+    requires_x64 = make_A is test_util.hermitian_matrix_eigvals_decaying
+    if requires_x64:
+        config.update("jax_enable_x64", True)
+    n = 1_000
+    num_rep = 10
```
pnkraemer, line 135: I understand that reproducing the experiments is the clearest test we could run. However, 10 reps of linear algebra with 1000x1000 matrices is quite costly (and we do this in multiple experiments); I think it's a bit too costly for a unittest to do this regularly.

I think it would be nice to reduce this size. If we want to verify that the performance matches the experiments in the thesis, we could create a benchmarks directory and include it in the documentation, and handle things there.

This could be a separate PR though. I am okay with leaving it in for now, but should we create an issue to migrate these tests to dedicated benchmarks?

pnkraemer (commented):
Thanks for the PR, and sorry for the delay in responding. I left some comments.

They are quite open-ended, and I look forward to hearing your opinion on the suggestions!

matfree/stochtrace.py
sethaxen: The requirement that the full input and output pytree structures are identical is much stricter than what the math requires (for x(nys)trace, x(nys)diag, and xsymrownorm): that the operator is square, i.e. the number of leaves is identical. I wonder if requiring that the pytrees be identical is too constraining, requiring some users to rewrite a perfectly reasonable matvec. But for our pytree-valued integrands, we do need to decide what we return.

For x(sym)rownorm, it makes sense for the result to have the same pytree as the matvec result (because that's the structure any column of the operator would have). For x(nys)trace, it doesn't matter what we require, since we always return a scalar. For x(nys)diag, we do need to make a decision about whether we return the input pytree or the output.

Or if we think the only sensible usage of these integrands (except xrownorm) is for matvecs with identical input and output pytree structures, we could require that everywhere.

pnkraemer: I would've thought that pytree-inputs = pytree-outputs is the pytree-equivalent of a "square matrix", in the sense of a function mapping from vector space $V$ to $V$ again. It seems that this is required for a trace to be defined, at least according to [wikipedia](https://en.wikipedia.org/wiki/Trace_(linear_algebra)#Trace_of_a_linear_operator). I also think that's the only case in which the notion of a "diagonal" is unambiguous. But I might be missing something here...

> I wonder if requiring that the pytrees be identical is too constraining, requiring some users to rewrite a perfectly reasonable matvec.

Do you perhaps have an example in mind? I think that would help me understand the issue a bit better

matfree/stochtrace.py
sethaxen: I shared in https://github.com/pnkraemer/matfree/issues/284#issue-4726343094 some of my reasoning for this choice. Namely, that the only difference between the methods is how the `Q` factor is constructed, so almost all of the code is shared.

But I do agree that they are different estimators. While `xsymrownorm` performs half the matvecs per test vector, that extra matvec `xrownorm` does is subspace iteration and meanginfully improves the quality of the estimate. This is visible in the `step` example in https://github.com/pnkraemer/matfree/pull/285#issue-4736798730, and if we plot all experimental results with the number of test vectors on the x-axis, we see that `xymrownorm` is strictly worse on the same test vectors:
fig17-2_numsamples_full

So I agree that simply differentiating the two with a keyword that indicates a mathematical property of the operator is misleading; a user may want to use `xrownorm` even for a normal operator because of the extra subspace iteration.

Perhaps if this integrand was factored into a standalone internal helper that both call (i.e. something like `_leave_one_out_xrownorms_xsymrownorms_squared`), that would give us the best of both worlds. What do you think?

matfree/stochtrace.py
sethaxen: Yeah I think it makes sense to handle that outside of this PR, since it could require not-insignificant changes to the other integrands.

tests/test_stochtrace/test_leave_one_out/test_xrownorms_squared.py
sethaxen: > Do you think it would be clearer if we moved the `dtype` to the case declarations? We can use `testing.parametrize` on the cases as well.

I'm a little unclear how what you're proposing differs from what I'm doing here; could you provide an example?

> There is also the option of making the test-case outputs into a dedicated datastructure (eg a dataclass) with those 5 fields, in which case we dont have to remember what the length-4/5 tuple stands for :)

Would this be automatically unpacked into the arguments of the test method, or would it be a single argument we would then need to unpack ourselves?

matfree/stochtrace.py
pnkraemer: Sorry, I forgot to look at the issue before commenting... Thanks for clarifying! 

Yes, internally I have no problems with both using the same function, and I agree it makes sense to reuse as much code as possible. My main concern was the user-facing bit :)

tests/test_stochtrace/test_leave_one_out/test_xrownorms_squared.py
sethaxen: > I understand that reproducing the experiments is the clearest test we could run. However, 10 reps of linear algebra with 1000x1000 matrices is quite costly (and we do this in multiple experiments); I think it's a bit too costly for a unittest to do this regularly.

Agreed. We could also register a custom mark for slow tests and in the CI skip slow tests by default.

> This could be a separate PR though. I am okay with leaving it in for now, but should we create an issue to migrate these tests to dedicated benchmarks?

Yeah I can do this, and we can also discuss other ideas there.

matfree/stochtrace.py
sethaxen: > Do you perhaps have an example in mind? I think that would help me understand the issue a bit better

The only one that came to mind is something like a user-provided operator that computes the time-derivative of a state space, e.g. 

```python
def matvec(x):
    """Damped oscillator: maps state -> time derivative of state."""
    q = x["position"]
    v = x["velocity"]

    stiffness = 2.0
    damping = 0.3

    return {
        "d_position_dt": v,
        "d_velocity_dt": -stiffness * q - damping * v,
    }
```
One could argue that the time derivative *should* have the same pytree as the state space itself. Or one could ask why one would ever need the trace or diagonal of this operator. I don't know the answer to either of those.

I think requiring that input and output pytrees be identical for square operators is fine, so long as well-documented. If a user comes with an operator where trace/diag make sense but identical pytrees does not, then maybe that would change our minds.

tests/test_stochtrace/test_leave_one_out/test_xrownorms_squared.py
pnkraemer: > Would this be automatically unpacked into the arguments of the test method, or would it be a single argument we would then need to unpack ourselves?

I would expect a namedtuple to be unpacked automatically, but a dataclass wouldnt. I guess it's a matter of personal preference whether one writes
```
test_xrownorms_squared_reproduce_experiments(setup: XRowNormSetup):
if config.requires_x64:
   config.update("jax_enable_x64", True)
n = setup.n
num_rep = setup.num_rep
```
etc. I sometimes like to do this when the test-cases get too complicated (eg over at Probdiffeq, I regularly do it because covering all possible solver-parametrisations blows up pretty quickly). But I didnt want to indicate that it has to be done, just a thought :) 

> I'm a little unclear how what you're proposing differs from what I'm doing here; could you provide an example?

Sure! I mean writing:
```python
@testing.parametrize("dtype", [float, complex])
def cases_experiments_step_general(dtype):
    """Hermitian matrix with eigenvalues that are flat with a sudden drop, XRowNorm"""
    return test_util.hermitian_matrix_eigvals_step, 60, 1e-5, False, dtype
```
and then 
```python
@testing.parametrize_with_cases("make_A, num_samples, max_rel_err, is_normal, dtype", cases=".")
def test_xrownorms_squared_reproduce_experiments(
    make_A, num_samples, max_rel_err, is_normal, dtype
):
    pass
```
which would avoid this pytest-cases symptom that some arguments are parametrised jointly (all except dtype) and some others are parametrised individually (dtype). I typically find it easier to parse if a test function with parametrisation-decorators has either as many decorators as parameters, or exactly one decorator. 

(Sorry for the superficial comments, this is again nothing groundbreaking so feel free to ignore!)

tests/test_stochtrace/test_leave_one_out/test_xrownorms_squared.py
```diff
@@ -0,0 +1,153 @@
...
+    def matvec(v, A):
+        return {"fx": A @ v["x"]}
+
+    key_ests = prng.split(key, num_rep)
+    received = func.vmap(lambda key: estimate(matvec, key, A))(key_ests)["fx"]
+    max_rel_errs = np.array_max(np.abs(received - expected) / np.abs(expected), axis=1)
+    median_max_rel_err = np.median(max_rel_errs)
+    assert float(median_max_rel_err) < max_rel_err
+    if requires_x64:
+        config.update("jax_enable_x64", False)
```
pnkraemer: Nice! :) 


Should we add an "enable_x64" context manager to the backend? Seems it exists now: https://docs.jax.dev/en/latest/_autosummary/jax.enable_x64.html. (Last time I looked it didnt, but that was a while ago)

matfree/stochtrace.py
pnkraemer: I agree. I'll create an issue so I don't forget :)

tests/test_stochtrace/test_leave_one_out/test_xrownorms_squared.py
pnkraemer: Thanks! Then let's leave it as is for this PR but keep the note around :+1:

matfree/stochtrace.py
pnkraemer: Interesting that you chose this example, because I actually believe very strongly that positions and velocities should indeed have the same pytree structure :smile: (That's a critical part of Probdiffeq)

So let's just assume they are the same, mention this in the docs, and wait until someone opens an issue with a concrete example that would make us change our mind? I think that would be a sensible way forward. What do you think?

sethaxen: Agreed!

tests/test_stochtrace/test_leave_one_out/test_xrownorms_squared.py
sethaxen: Yeah that would make sense. Although, it's a bit annoying that it requires changing the indentation for the rest of the method wrt the other experiment. Also, how would we only conditionally use that context manager?

pnkraemer: Right; sorry for the misleading suggestion. Let's scratch that then...

Two other small things:
- Should we make the "requires_x64" a part of the test-case because then, the decaying eigvals testcase can "decide themselves" that they want double precision and the test itself doesnt have to hard-code this "if make_A is decaying" check? 
- Now, that I look at it again, I think
```python
if requires_x64:
    config.update("jax_enable_x64", True)
```
could be `config.update("jax_enable_x64", requires_x64)`. I am surprised the linter didn't catch that...

sethaxen:
Okay, I think I've made all of the requested changes within the scope of this PR.

One question RE API. In https://github.com/pnkraemer/matfree/pull/285#discussion_r3491091473 you suggested using the names `leave_one_out_xrownorm_squared` and `leave_one_out_xsymrownorm_squared` (i.e. singular `...rownorm_squared`), where we currently user `...rownorms_squared`. Was this a typo, or would you like to adopt the singular? If the latter (which uses the algorithm names), should we also change the Monte Carlo variant to `monte_carlo_rownorm_squared`?

matfree/backend/func.py
```diff
@@ -46,6 +46,18 @@ def linear_transpose(func, *primals):
...
+def linear_adjoint(func, *primals):
+    transpose = jax.linear_transpose(func, *primals)
+
+    def tree_conj(x):
+        return jax.tree.map(lambda x: x.conj(), x)
+
+    def adjoint(*primal_results):
+        return tree_conj(transpose(*tree_conj(primal_results)))
+
+    return adjoint
```
sethaxen, line 58: Should we comprehensively test this somewhere? Currently our stochtrace tests only test the array argument version.

pnkraemer: I think it's fine without. Its correctness is verified through the stochtrace tests and for a backend-function, that should be plenty.

pnkraemer:
> One question RE API. In https://github.com/pnkraemer/matfree/pull/285#discussion_r3491091473 you suggested using the names leave_one_out_xrownorm_squared and leave_one_out_xsymrownorm_squared (i.e. singular ...rownorm_squared), where we currently user ...rownorms_squared. Was this a typo, or would you like to adopt the singular? If the latter (which uses the algorithm names), should we also change the Monte Carlo variant to monte_carlo_rownorm_squared?

Thanks for flagging this! Honestly, I am not sure; the thesis calls the algorithm `xrownorm`, but in plain English I would say we compute the row-norm*s* because otherwise it sounds like we compute the norm of a single row (which raises the follow-up question of "which row?").

The longer I think about this, the more I lean towards the plural. But I see cases for both; what do you think?

pnkraemer (commented):
Looks good, thanks for the edits! 

As soon as the naming discussion is complete, I'll approve and merge :)

sethaxen:
> The longer I think about this, the more I lean towards the plural. But I see cases for both; what do you think?

I also prefer plural for the API here, as it makes the usage of this function clearer.

pnkraemer (approved):

PR #288, opened 2026-07-15
Fix Hermitian Lanczos for complex matrices

rahulgaur104:
Hi @pnkraemer,
Thank you for the nice package. I have been using matfree to solve a Hermitian eigenvalue stability problem. However, I realized that with a complex matrix, matfree requires some minor modifications for it to work correctly. 

## Summary
- use the conjugating inner product in Arnoldi/Hessenberg projections and norms
- symmetrize tridiagonal matrices with the Hermitian transpose
- update the direct Lanczos path and tests for complex Hermitian matrices

## Tests
- `python -m pytest` (415 passed)

### DIFF
--- a/matfree/decomp.py
+++ b/matfree/decomp.py
@@ -130,7 +130,7 @@ def _tridiag_reortho_full(num_matvecs: int, /, *, custom_vjp: bool, materialize:
     def estimate(matvec, vec, *params):
         Q, H, v, norm = alg(matvec, vec, *params)
 
-        T = 0.5 * (H + H.T)
+        T = 0.5 * (H + H.T.conj())
         diags = linalg.diagonal(T, offset=0)
         offdiags = linalg.diagonal(T, offset=1)
 
@@ -219,9 +219,9 @@ def _tridiag_reortho_none(num_matvecs: int, /, *, custom_vjp: bool, materialize:
 
 def _tridiag_forward(matvec, num_matvecs, vec, *params):
     # Pre-allocate
-    vectors = np.zeros((num_matvecs + 1, len(vec)))
-    offdiags = np.zeros((num_matvecs,))
-    diags = np.zeros((num_matvecs,))
+    vectors = np.zeros((num_matvecs + 1, len(vec)), dtype=vec.dtype)
+    offdiags = np.zeros((num_matvecs,), dtype=vec.dtype)
+    diags = np.zeros((num_matvecs,), dtype=vec.dtype)
 
     # Normalize (not all Lanczos implementations do that)
     v0 = vec / linalg.vector_norm(vec)
@@ -261,7 +261,7 @@ def _tridiag_fwd_init(matvec, vec, *params):
     for x_{k+1}, a_k, and b_k, using
     orthogonality of the x_k.
     """
-    a = vec @ (matvec(vec, *params))
+    a = linalg.vdot(vec, matvec(vec, *params))
     r = (matvec(vec, *params)) - a * vec
     b = linalg.vector_norm(r)
     x = r / b
@@ -285,7 +285,7 @@ def _tridiag_fwd_step(matvec, params, i, val):
 
 def _tridiag_fwd_step_apply(matvec, vec, b, vec_previous, *params):
     """Apply Lanczos' recurrence."""
-    a = vec @ (matvec(vec, *params))
+    a = linalg.vdot(vec, matvec(vec, *params))
     r = matvec(vec, *params) - a * vec - b * vec_previous
     b = linalg.vector_norm(r)
     x = r / b
@@ -353,8 +353,8 @@ def hessenberg(
 ):
     r"""Construct a **Hessenberg-factorisation** via the Arnoldi iteration.
 
-    Factorise $A \approx Q H Q^\top$, where $Q$ is orthogonal and $H$ is upper Hessenberg.
-    Works for **arbitrary square matrices**. Does not support complex-valued matrices.
+    Factorise $A \approx Q H Q^H$, where $Q$ is orthonormal and $H$ is upper Hessenberg.
+    Works for **arbitrary square matrices** including complex-valued matrices.
 
     Setting `custom_vjp` to `True` implies using efficient, numerically stable
     gradients of the Arnoldi iteration which was proposed by Krämer et al. (2024).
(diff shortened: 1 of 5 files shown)


rahulgaur104:
Closing this fork-based PR; I will try pushing a branch directly to the upstream repository instead.

rahulgaur104:
Opening this fork-based PR, as I don't have permissions to create a separate branch.

matfree/decomp.py
```diff
@@ -353,8 +353,8 @@ def hessenberg(
 ):
     r"""Construct a **Hessenberg-factorisation** via the Arnoldi iteration.
 
-    Factorise $A \approx Q H Q^\top$, where $Q$ is orthogonal and $H$ is upper Hessenberg.
-    Works for **arbitrary square matrices**. Does not support complex-valued matrices.
+    Factorise $A \approx Q H Q^H$, where $Q$ is orthonormal and $H$ is upper Hessenberg.
+    Works for **arbitrary square matrices** including complex-valued matrices.
```
pnkraemer: Could you leave a note in the docstring that while the forward pass now supports complex matrices, the backward pass doesn't? (Ideally, an error message in the adjoint code with a corresponding test, but I would be okay with a note in the docstring for now)

For context, this requires a slightly larger fix than changing transposes to conjugate transposes and should be handled separately. However, it would be nice to communicate this gap to a user.

matfree/decomp.py
```diff
@@ -261,7 +261,7 @@ def _tridiag_fwd_init(matvec, vec, *params):
     for x_{k+1}, a_k, and b_k, using
     orthogonality of the x_k.
     """
-    a = vec @ (matvec(vec, *params))
+    a = linalg.vdot(vec, matvec(vec, *params))
```
pnkraemer, line 266: Just so you know: Currently, matfree uses linalg.vdot and linalg.inner relatively inconsistently, and I made a note (https://github.com/pnkraemer/matfree/issues/289) about fixing this in the midterm. 

I am not saying your use of vdot is wrong (not at all!), but be warned that it might be refactored in the near future :)

tests/test_decomp/test_tridiag_sym.py
```diff
@@ -23,7 +26,7 @@ def matvec(s, p):
 
     # Reconstruct the original matrix from the full-num_matvecs approximation
     # Q shape is (k, n) -- rows are Krylov vectors
-    matrix_reconstructed = Q.T @ T @ Q
+    matrix_reconstructed = Q.T @ T @ Q.conj()
```
pnkraemer, line 29: Why is it `Q.T @ T  @ Q.conj()` and not `Q.T.conj() @ T @ Q`?

tests/test_eig/test_eigh_partial.py
```diff
@@ -20,6 +20,23 @@ def test_equal_to_linalg_eigh(nrows):
     assert np.allclose(vals, S)
 
 
+@testing.parametrize("nrows", [48])
+@testing.parametrize("num_matvecs", [24])
+def test_ritz_values_stay_bounded_for_complex_hermitian(nrows, num_matvecs):
```
pnkraemer: Is there a specific failure case this test catches? As is, it feels a bit redundant given the other tests but I might be missing something.

pnkraemer (commented):
Thanks a lot for the PR! Really nice to make Lanczos and Arnoldi work for complex matrices.

I left some comments in the diff, would be nice to get them sorted before merging - but nothing major!

matfree/decomp.py
rahulgaur104: Thanks for the heads-up. I’ll keep `vdot` here for now because the Lanczos projection needs the conjugating inner product for complex inputs, but I agree this can be revisited as part of the broader `inner`/`vdot` cleanup in #289.

tests/test_decomp/test_tridiag_sym.py
rahulgaur104: The reason is that the returned `Q` stores Krylov vectors as rows, so `Q_col = Q.T`. The usual column-basis reconstruction `Q_col @ T @ Q_col.conj().T` therefore becomes `Q.T @ T @ Q.conj()`. I added a short comment in the test to make this convention explicit.

tests/test_eig/test_eigh_partial.py
rahulgaur104: Yes, this catches the user-facing failure mode from the complex Hermitian bug. On `origin/main`, this case produced a largest Ritz value above the true largest eigenvalue, while it passes with this PR. I renamed the test to mark it as a regression test.

matfree/decomp.py
rahulgaur104: Done, I added a note.

rahulgaur104:
@pnkraemer 
I think this is ready for review.

rahulgaur104:
Unrelated but I used your package to write a dense and matrix-free shift invert solver [here](https://arxiv.org/abs/2608.01750) (page number 11). 
If you want, we can add a shift invert solver to matfree.

tests/test_eig/test_eigh_partial.py
pnkraemer: Thanks for following up! 

Does this regression test catch something that is not caught by the dtype=complex additions to the other tests? I agree it's good to verify that the changes lead to improved behaviour, but I would expect that the other test modifications in this PR would've also failed without the changes to the src. Or am I missing something?

pnkraemer:
> Unrelated but I used your package to write a dense and matrix-free shift invert solver [here](https://arxiv.org/abs/2608.01750) (page number 11). If you want, we can add a shift invert solver to matfree.

Super nice! Let's discuss this in the issue thread?

tests/test_eig/test_eigh_partial.py
rahulgaur104: You're right. This test catches the user-facing symptom, but it does not cover anything else beyond the `dtype=complex` additions to the tridiagonalization tests. I’ll remove it to keep the PR focused.

rahulgaur104: Removed.

pnkraemer: Thanks!

pnkraemer (approved):
Thanks a lot for that contribution! :)

